PR #5, opened 2023-07-10
Insert Table of Contents if tag exists

jloeser:
There is https://docs.gitlab.com/ee/user/markdown.html#table-of-contents describing `[[_TOC_]]` or `[TOC]` tag. Let's replace these tags by corresponding Confluence XHTML.

### DIFF
--- a/md2conf/converter.py
+++ b/md2conf/converter.py
@@ -177,6 +177,17 @@ class ConfluenceStorageFormatConverter(NodeVisitor):
             AC("plain-text-body", ET.CDATA(content)),
         )
 
+    def _transform_toc(self, code: ET.Element) -> ET.Element:
+        return AC(
+            "structured-macro",
+            {
+                ET.QName(namespaces["ac"], "name"): "toc",
+                ET.QName(namespaces["ac"], "schema-version"): "1",
+            },
+            AC("parameter", {ET.QName(namespaces["ac"], "name"): "outline"}, "clear"),
+            AC("parameter", {ET.QName(namespaces["ac"], "name"): "style"}, "default"),
+        )
+
     def transform(self, child: ET.Element) -> Optional[ET.Element]:
         # normalize line breaks to regular space in element text
         if child.text:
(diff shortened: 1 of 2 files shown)


md2conf/converter.py
```diff
@@ -273,6 +273,20 @@ def __init__(self, path: str, options: ConfluenceDocumentOptions) -> None:
...
+            '<ac:parameter ac:name="style">default</ac:parameter>',
+            '</ac:structured-macro>',
+        ]
+
+        # replace Table of Contents tag
+        html = re.sub(
+           r"(<p>(\[(TOC)\]|\[\[<em>(TOC)</em>\]\])\s*</p>)",
+           "\n".join(toc),
+           html
+        )
```
hunyadi: That's a simple but cool feature, thanks for contributing!

Most of the transformations are implemented in the class `ConfluenceStorageFormatConverter`, which is operating on the element tree constructed by `lxml.etree`. Operating on the element tree with a node visitor is more efficient, less fragile and easier to understand that doing a second pass on the generated output with a regular expression. Could you move the injection of *Table of Contents* into `transform` in `ConfluenceStorageFormatConverter`?

jloeser: Yeah, I was already thinking of this but well... this was the fastest way. I will give your suggestion a try.

md2conf/converter.py
```diff
@@ -190,6 +201,11 @@ def transform(self, child: ET.Element) -> Optional[ET.Element]:
         if child.tag == "p" and len(child) == 1 and child[0].tag == "img":
             return self._transform_image(child[0])
 
+        # <p>[[_TOC_]]</p>
+        # <p>[TOC]</p>
+        elif child.tag == "p" and "".join(child.itertext()).startswith(("[[TOC]]", "[TOC]")):
```
hunyadi: Thanks for the update, I hope you agree that this looks much nicer. ;)

Documentation for GitLab says:
> Add either the `[[_TOC_]]` or `[TOC]` tag on its own line

Both of your code comments (`<p>[TOC]</p>` and `<p>[[_TOC_]]</p>`) seem to confirm. This means we shouldn't need the call to `startswith`, do we? We could use string equality. I would like us to be as strict as possible to avoid picking up something that looks like a table of contents but in fact is not (e.g. someone writing TOC in the body of text for *terms of contract*).

hunyadi (approved):

PR #6, opened 2023-07-11
update tool for directory usage

dgeorges:
Update tool to sync an entire directory to confluence. 

The sync of a directory also converts relative links with in the pages.

For example when converting a page site/references/topic/README.md with the following content:

```
<!-- confluence-page-id: 118981598 -->
<!-- confluence-space-key: Space -->
# A Title

[relative_link_example](../how_to_guides/do_something.md)
[sibling_page_example](sister.md)
[anchor_example](#More-Info)

```
Will be converted into

```
[relative_link_example](https://${CONFLUENCE_DOMAIN}/wiki/spaces/${do_something.md[space_key]}/pages/${do_something.md[page_id]})
[sibling_page_example](https://${CONFLUENCE_DOMAIN}/wiki/spaces/${sister_page.md[space_key]}/pages/${sister_page.md[page_id]})
[anchor_example](https://${CONFLUENCE_DOMAIN}/wiki/spaces/118981598/pages/Space/A+Title#More-Info)
```

** Note relative links only get converted if they point to resources in the directory which you are syncing


md2conf/api.py
```diff
@@ -328,6 +328,8 @@ def update_page(self, page_id: str, new_content: str) -> None:
         page = self.get_page(page_id)
 
         try:
+            # sanitize_confluence throws an error for unknown characters. Which if not fix will always result in a page update.
+            # as old_content != new_content, since confluence will not save unknown characters.
```
hunyadi: From what I recall, this was with UTF-8 characters that are not in ASCII, which are perfectly fine in HTML (and Confluence XHTML) but appear to cause an issue with the `lxml` parser. I think if we do a first pass on the document, and replace these characters, we can avoid the infinite update loop.

md2conf/application.py
```diff
@@ -15,16 +23,70 @@ def update_document(
     api.update_page(document.page_id, document.xhtml())
 
 
+
+def synchronize_pages(api: ConfluenceSession, path: str, options: ConfluenceDocumentOptions) -> None:
```
hunyadi: We might want to wrap this (and other related functions) in a new class so that we pass the parameters once (in the new class `__init__`) and then access them via `self` in member functions like `synchronize_pages`, `synchronize_page`, etc., and don't need to propagate them below.

md2conf/application.py
```diff
@@ -15,16 +23,70 @@ def update_document(
...
+    if os.path.isdir(path):
+        synchronize_directory(api, path, options)
+    elif os.path.isfile(path):
+        synchronize_page(api, path, options)
+    else:
+        raise ValueError(f"expected: valid file of directory path. got: {path}")
```
hunyadi: Likely a typo, should read:

> ValueError(f"expected: valid file or directory path; got: {path}")

md2conf/application.py
```diff
@@ -15,16 +23,70 @@ def update_document(
...
+
+
 def synchronize_page(
-    api: ConfluenceSession, path: str, options: ConfluenceDocumentOptions
+    api: ConfluenceSession, path: str, options: ConfluenceDocumentOptions, page_metadata: dict[str, str] = dict()
```
hunyadi: This package is marked as Python 3.8-compatible in [PyPI](https://pypi.org/project/markdown-to-confluence/). As such, I am afraid we must stick to `typing.Dict` for annotating dictionary types.

md2conf/application.py
```diff
@@ -15,16 +23,70 @@ def update_document(
...
+def synchronize_directory(api: ConfluenceSession, dir: str, options: ConfluenceDocumentOptions) -> None:
+    page_metadata = dict()
+    LOGGER.info(f"synchronize_directory: {dir}")
+
+    # Step 1: build index of all page metadata
+    for root, directories, files in os.walk(dir):
+        for filename in files:
+            # Extract the file extension from the path
+            file_extension = os.path.splitext(filename)[1]
+            abs_path = os.path.abspath(os.path.join(os.path.abspath(root), filename))
```
hunyadi: I am not sure I fully understand the purpose of the outer call to `os.path.abspath`, looks like we don't need it at all.

md2conf/application.py
```diff
@@ -15,16 +23,70 @@ def update_document(
...
+            if file_extension.lower() == ".md":
+                # Open file
+                with open(abs_path, "r") as f:
+                    # document = markdown_to_html(f.read())
```
hunyadi: If we no longer need this call, can you remove this commented out line?

md2conf/converter.py
```diff
@@ -250,49 +306,57 @@ class ConfluenceDocument:
...
+        self.space_key, html = extract_value(
             r"<!--\s+confluence-space-key:\s*(\w+)\s+-->", html
         )
 
         # parse Markdown document
         if self.options.generated_by:
             content = [
                 '<ac:structured-macro ac:name="info" ac:schema-version="1">',
-                "<ac:rich-text-body><p>This page has been generated with a tool.</p></ac:rich-text-body>",
+                '<ac:rich-text-body><p> Do Not Edit In Confluence: This page is being periodically imported from the <a href="https://infra-wiki.onemedical.io/">infra-wiki</a> with a tool.</p></ac:rich-text-body>',
```
hunyadi: I think you should revert to the original text here. However, we can think of options how to configure what appears here.

md2conf/converter.py
```diff
@@ -305,6 +369,17 @@ def sanitize_confluence(html: str) -> str:
     return _content_to_string(root)
 
 
+def _content_title(root: ET.Element) -> str:
+    xml = ET.tostring(root, encoding="utf8", method="xml").decode("utf8")
+    m = re.match(r".*<h1>(.*)</h1>.*", xml, re.DOTALL)
+
+    # if the contents of the page are empty
+    if m == None:
```
hunyadi: While it's somewhat of a personal taste question, I prefer to use `m is None` since `None` is a unique object in Python.

md2conf/application.py
```diff
@@ -15,16 +23,70 @@ def update_document(
...
 def synchronize_page(
-    api: ConfluenceSession, path: str, options: ConfluenceDocumentOptions
+    api: ConfluenceSession, path: str, options: ConfluenceDocumentOptions, page_metadata: dict[str, str] = dict()
 ) -> None:
     page_path = os.path.abspath(path)
     base_path = os.path.dirname(page_path)
 
-    document = ConfluenceDocument(path, options)
+    LOGGER.info(f"synchronize_page: {page_path}")
+    document = ConfluenceDocument(path, options, page_metadata)
```
hunyadi: Maybe it's an oversight on my end but does this line up with the type annotation above and the declared signature of `ConfluenceDocument`? Can you run [mypy](https://mypy.readthedocs.io/en/stable/) on the source to check?

md2conf/converter.py
jloeser: I have already thought of that and am already preparing if you don't mind.

hunyadi: To the contrary, I would most appreciate it. As for the Python code, the text could be passed as another configuration option. As for the command-line interface to `md2conf`, the text should likely come from a configuration file, preferably JSON, because it's a longer piece, and putting that all as a command-line parameter is likely not that practical. We don't have an option to pass a configuration file to `md2conf` at the moment, and then parse it, but that could be added later. Both of these, in my opinion, are out of scope for this pull request, and should be done in a separate one.

However, we need to revert adding a reference to an internal wiki site here such that I can merge this pull request.

jloeser: ... let's discuss this here https://github.com/hunyadi/md2conf/pull/7

md2conf/application.py
dgeorges: Makes sense!

md2conf/application.py
dgeorges: You are correct. fixed!

md2conf/api.py
dgeorges: Some of our markdown had characters such as “ instead of ", likely the cause of copy paste. What would happen is that content of first upload would be published as `&rsquo;` in confluence. And then upon running the tool again, sanitize would throw. The result would be to publish the page even when there were no changes made.

hunyadi: Yeah, this is because `lxml` by default doesn't recognize HTML entities (character escapes like `&rsquo;`). I gave this a shot earlier but it turned out to require more than the 30 minutes I had at that time. Once HTML entities are properly handled, `sanitize` shouldn't have any issues processing documents with entities in them.

md2conf/converter.py
```diff
@@ -115,20 +125,64 @@ def _change_ext(path: str, target_ext: str) -> str:
...
+        self.path = path
+        self.base_path = os.path.dirname(path)
         self.links = []
         self.images = []
+        self.page_metadata = page_metadata
 
     def _transform_link(self, anchor: ET.Element) -> ET.Element:
         url = anchor.attrib["href"]
         if not is_absolute_url(url):
+            LOGGER.debug(f"not is_absolute_url = {url} from = {self.path}")
```
hunyadi: I am not 100% sure I would understand what this log message means if I encountered it in a log file. Would it be better to say:
```
LOGGER.debug(f"found link {url} relative to {self.path}")
```

md2conf/converter.py
```diff
@@ -115,20 +125,64 @@ def _change_ext(path: str, target_ext: str) -> str:
...
             self.links.append(url)
+            # Convert the relative href to absolute based on relative url the base path value then
+            # look up the absolute path in the page metadata dictionary to discover
+            # the relative path within confluence that should be used
+            abs_path = os.path.abspath(
+                os.path.join(os.path.abspath(self.base_path), url)
+            )
```
hunyadi: Like in a previous case, the outer call to `os.path.abspath` doesn't seem to be necessary. (Unless, of course, we want to ensure that `url` doesn't walk us out of the directory root, e.g. with `../../../path/to/dir`.)

md2conf/converter.py
```diff
@@ -115,20 +125,64 @@ def _change_ext(path: str, target_ext: str) -> str:
...
+
+            # Transform relative urls like "#page_anchor" to "$pagepath#page_anchor"
+            if url.startswith("#"):
+                abs_path = self.path + url
+
+            # The path prior to the page anchor if it exists
+            page_path = abs_path.split("#")[0]
```
hunyadi: Have you considered the Python module [urllib.parse](https://docs.python.org/3/library/urllib.parse.html)? It may be a more straightforward (and future-proof) solution to split URLs and then re-construct URLs from components than string manipulation, which can have its caveats.

md2conf/converter.py
```diff
@@ -115,20 +125,64 @@ def _change_ext(path: str, target_ext: str) -> str:
...
+
+            # lookup page metadata using the page path (without the achor)
+            link_metadata = self.page_metadata.get(page_path)
+            relative_url = None
+            if link_metadata:
+                LOGGER.debug(f"found page {page_path} with metadata: {link_metadata}")
+                confluence_page_id = link_metadata.page_id
+                confluence_space_key = link_metadata.space_key
+                relative_url = f"https://onemedical.atlassian.net/wiki/spaces/{confluence_space_key}/pages/{confluence_page_id}"
```
hunyadi: This seems something that wouldn't be portable due to the hard-wired base URL.

md2conf/api.py
dgeorges: I am happy to remove the comment in favor of an issue if you like.

md2conf/converter.py
dgeorges: Without the outer abspath you can get paths like
`/Users/name/site/./folder/README.md`

dgeorges:
Thanks for the prompt reviews. I apologies for the sloppiness. I was intending to merge into my fork first, not realizing I was on the upstream project. I be a bit occupied for the rest of the week, but should be able to address everything early next week.

md2conf/api.py
hunyadi: This issue has been addressed on the mainline branch `master` by adding an entity definition document `entity.dtd`.

dgeorges: Thanks, I've resync'd and it works well!

dgeorges:
> Thanks for the prompt reviews. I apologies for the sloppiness. I was intending to merge into my fork first, not realizing I was on the upstream project. I will be a bit occupied for the rest of the week, but should be able to address everything early next week.

Ok, I think I addressed everything. PTAL

md2conf/converter.py
```diff
@@ -142,20 +153,60 @@ def _change_ext(path: str, target_ext: str) -> str:
...
+            # lookup page metadata using the page path
+            link_metadata = self.page_metadata.get(abs_path)
+            transformed_url = None
+            if link_metadata:
+                LOGGER.debug(f"found page {abs_path} with metadata: {link_metadata}")
+                transformed_url = f"https://{link_metadata.domain}/wiki/spaces/{link_metadata.space_key}/pages/{link_metadata.page_id}"
+
+                if relative_url.fragment:
+                    # confluence url's use `+` instead of spaces for the page title part
+                    confluence_page_title = link_metadata.title.replace(" ", "+")
```
hunyadi, line 197: Instead of ad-hoc string manipulation, I would use the built-in Python package `urllib.parse`. Specifically, this package has [urlencode](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode), which not only replaces spaces with `+` but also escapes other special characters. Also, I would suggest [urlunparse](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlunparse) to construct URLs from parts, not piece them together as strings, which can be error prone.

hunyadi (commented):
Thanks for the updates, they are awesome! I might need to make small adjustments to the code but in general it's great! Specifically, I would like to extend unit tests. Currently, we are testing the conversion of a flat file `example.md` (see unit test `test_synchronize_page`). However, since this pull request introduces conversion of directory hierarchies, it would be great to have a sample hierarchy of Markdown documents, and have a test to publish them to Confluence. If you have some time to extend `test_api.py`, it would be extremely helpful. However, the pull request is great as it is, and I can also take the responsibility to update unit tests later.

hunyadi (approved):

dgeorges:
> test_api.py

Thanks! I would love to add tests for the 3 types of hyperlinks mentioned in the PR description, plus an absolute url. I see what I can do for that, but I might wait till you fully merged the PR, this week is already a bit busy anyways.

PR #7, opened 2023-07-12
Allow user-defined generated-by prompt text

jloeser:


README.md
```diff
@@ -58,6 +58,15 @@ Each Markdown file is associated with a Confluence wiki page with a Markdown com
 
 The above tells the tool to synchronize the Markdown file with the given Confluence page ID. This implies that the Confluence wiki page must exist such that it has an ID. The comment can be placed anywhere in the source file.
 
+### Provide generated-by prompt text in Markdown file
+
+Provide generated-by prompt text in Markdown file via the following optional tag:
+
+```markdown
+<!-- generated-by: Do not edit! Origin can be found <a href="https://git.example.com/project">here</a>. -->
```
hunyadi, line 66: I am afraid I don't like this approach. At Instructure (the company I work for), we typically have a single generic text for all documents, e.g. all documents exported from the same Git repository have the same label that points to the Git repository root. I believe this generalizes to other companies and use cases, you don't typically want to set this on a per-document basis. As such, I believe a separate configuration file would be more useful; you configure the text once, and apply the text to a bunch of Markdown documents.

README.md
```diff
@@ -58,6 +58,15 @@ Each Markdown file is associated with a Confluence wiki page with a Markdown com
...
+```
+Alternatively, use the `--generate-by GENERATED_BY` option. The tag takes precedence.
```
hunyadi: If you have a short and simple text, this is definitely a feasible approach. I like this much better. There is likely a typo in the parameter name, it should probably read: `generated_by`.

README.md
jloeser: Just to get it right: you think of a global configuration file (e.g. `~/.config/md2conf/config.yaml`) containing something in the sense:
```
---
generated-by: "This is my generic text for all documents."
```
I explicitly would like to define it per-document as I have several Markdown files laying around in different repositories where I would like to have a pointer (link) to these repositories in my generated Confluence pages.

Moreover, I don't think this is an either-or question, right? Why not providing all possibilities? One could define it per-document, or in a configuration file or via parameter. Precedence could be: parameter > per-document > global configuration file. Remember, this `generated-by` tag is optional.

README.md
jloeser: > If you have a short and simple text, this is definitely a feasible approach.

I was thinking of also parsing the prompt text but hey, it's intended to be a one-liner hint only.

> I like this much better.

It's already functional if you do not provide any `generated_by` tag. Again, this `generated-by` tag is optional.

And yes, typo.

README.md
hunyadi: Yeah, that's a fair point. Let's keep the per-document tag too.

README.md
jloeser: Typo is fixed.

README.md
jloeser: Still my question: do you plan to have this kind of configuration file for md2conf (this should happen in a separate PR anyway)?

hunyadi: Yes, I can add such a configuration file later. (I am currently busy with another project.)

jloeser: Alright. Then feel free to merge this PR.

md2conf/converter.py
```diff
@@ -289,11 +289,20 @@ def __init__(self, path: str, options: ConfluenceDocumentOptions) -> None:
...
         )
 
+        # extract 'generated-by' tag text
+        generated_by_tag, html = _extract_value(
+            r"<!--\s+generated-by:\s*(.*)\s+-->", html
+        )
+
         # parse Markdown document
         if self.options.generated_by:
+            generated_by = self.options.generated_by
```
hunyadi, line 299: Can you double-check if the command-line option `--no-generated-by` still works after these changes?

jloeser: I did and it works. If `--no-generated-by` is provided (and only then), `self.options.generated_by` becomes `False` because of `action=store_false`. Otherwise it is always a string, either default or user-defined, so this if-clause becomes `True`.

hunyadi (approved):

PR #22, opened 2023-12-07
Feature: Use 'Tiny link' instead of space/page link

Joffreybvn:
Hello,

The generated links do not work for my Confluence (I work on a self-hosted version): The _confluence/space/ID/page/PAGE_ID_ returns a 404... Looking into Confluence's docs, there is a way to [generate "Tiny links" based on a page id](https://confluence.atlassian.com/confkb/how-to-programmatically-generate-the-tiny-link-of-a-confluence-page-956713432.html).

This PR is a proposal that implements this conversion, so that generated links are tiny links - _which I guess should work in all cases._

By the way, great lib !

### DIFF
--- a/md2conf/converter.py
+++ b/md2conf/converter.py
@@ -1,4 +1,6 @@
 import importlib.resources as resources
+import base64
+import struct
 import logging
 import os.path
 import pathlib
@@ -53,6 +55,17 @@ def markdown_to_html(content: str) -> str:
             "sane_lists",
         ],
     )
+
+def page_id_to_tiny_link(page_id: int) -> str:
+    """Convert confluence page id into tiny link id."""
+    packed_data = struct.pack("Q", page_id)
+    return base64.b64encode(packed_data).decode('ascii')
+
+def tiny_link_to_page_id(tiny_id: str) -> int:
+    """Convert confluence tiny link id into page id."""
+    padded = tiny_id.ljust(11, "A") + "="
+    decoded_id = base64.b64decode(padded)
+    return struct.unpack("Q", decoded_id)[0]
 
 
 def _elements_from_strings(dtd_path: pathlib.Path, items: List[str]) -> ET._Element:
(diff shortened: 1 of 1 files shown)


md2conf/converter.py
```diff
@@ -297,10 +310,11 @@ def _transform_link(self, anchor: ET._Element) -> None:
         )
         self.links.append(url)
 
+        tiny_link_id: str = page_id_to_tiny_link(int(link_metadata.page_id))
         components = ParseResult(
             scheme="https",
             netloc=link_metadata.domain,
-            path=f"{link_metadata.base_path}spaces/{link_metadata.space_key}/pages/{link_metadata.page_id}/{link_metadata.title}",
+            path=f"{link_metadata.base_path}x/{tiny_link_id}",
```
hunyadi, line 317: I am afraid the logic may be more complex than this.

If I follow the instructions on the page [How to programmatically generate the tiny link of a Confluence page](https://confluence.atlassian.com/confkb/how-to-programmatically-generate-the-tiny-link-of-a-confluence-page-956713432.html), and click the *Share* button to generate a tiny URL, then I get a very different URL in my Confluence environment than what the function `page_id_to_tiny_link` returns. In fact, the generated URL is different each time I load the page and click *Share*, so I am assuming a lookup table is being used.

I would also make tiny links a configurable setting and not force this on all users, since tiny URLs appear to be a redirection feature, and the longer structured URLs seem to be the standard way that Confluence uses for routing.

hunyadi:
Unfortunately, these types of tiny links are not universal. For example, they are explicitly disabled in our managed Confluence, and only the "traditional" URLs are permitted, which *md2conf* already supports, and another type of shortened link (without the `x/` prefix), which is volatile. We must make this setting configurable (opt-in) in order to add it to the utility such that we don't break existing publish pipelines for *md2conf* users.

hunyadi:
I am closing this pull request due to inactivity.

PR #23, opened 2023-12-26
Add support for creating new pages in Confluence

GMous:
Added support for automatically creating new pages in Confluence when the processed Markdown file lacks an appropriate Page Id. When such files are found, the file name is assumed to be the page title. The system then tries to retrieve a page with this title from Confluence. If the page exists, the page id is written to the file. If the page is not found, a new page is created using the root page specified in the command line arguments, and the page id is written to the file.

### DIFF
--- a/md2conf/__main__.py
+++ b/md2conf/__main__.py
@@ -59,6 +59,11 @@ parser.add_argument(
     help="Use this option to set the log verbosity.",
 )
 parser.add_argument(
+    "-r",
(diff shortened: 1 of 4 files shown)


hunyadi (approved):
Thanks for contributing this feature! This is something I have wanted to do but never had the time. I have changed the target branch to allow me to make minor stylistic and technical changes, and run a round of testing before your contribution is merged into the mainline branch.

PR #32, opened 2024-04-18
Visit children before transforming nodes (post-order DFS)

AdamGagorik:
I noticed links were not working inside admonitions.  This is because we never visit them in the node visitor.  I think it should be changed from pre-order depth first search to post-order depth first search.  This is just something for you to consider, I don't know the full ramifications of this change without more testing.  With the change, I see links inside admonitions are working, but I don't know if it broke something else.

### DIFF
--- a/md2conf/converter.py
+++ b/md2conf/converter.py
@@ -197,11 +197,10 @@ class NodeVisitor:
 
         for index in range(len(node)):
             source = node[index]
+            self.visit(source)
             target = self.transform(source)
             if target is not None:
                 node[index] = target
-            else:
-                self.visit(source)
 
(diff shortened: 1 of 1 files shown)


AdamGagorik:
For sure this PR breaks code blocks (they end early), so maybe you should not take this change.

hunyadi:
A more robust approach, in my opinion, is to trigger the node visitor for descendant elements of the admonition, which confines the behavior to that element type, and doesn't interfere with other block types, e.g. code blocks. I have added an example in which links are properly transformed in the context of an admonition using this approach. Changes are pushed to `master`.

PR #33, opened 2024-04-18
Use a different URL for links

AdamGagorik:
Links were not working for me.  It could just be the Confluence server I have to deal with.  I do not know if it would work universally, so feel free to disregard this change.

I followed [this page](https://confluence.atlassian.com/confkb/the-differences-between-various-url-formats-for-a-confluence-page-278692715.html) to build links that would work for me.

- Made sure suffix (.md) was not in the generated link
- Made the link of the form `https://confluence.example.com/display/SPACEKEY/Page#Fragment`

Maybe this should be a configuration?

### DIFF
--- a/md2conf/converter.py
+++ b/md2conf/converter.py
@@ -280,7 +280,6 @@ class ConfluenceStorageFormatConverter(NodeVisitor):
             else:
                 raise DocumentError(msg)
 
-        relative_path = os.path.relpath(absolute_path, self.base_path)
         relative_path = os.path.splitext(os.path.relpath(absolute_path, self.base_path))[0]
 
         link_metadata = self.page_metadata.get(absolute_path)
--- a/md2conf/converter.py
+++ b/md2conf/converter.py
@@ -301,7 +301,6 @@ class ConfluenceStorageFormatConverter(NodeVisitor):
         components = ParseResult(
             scheme="https",
             netloc=link_metadata.domain,
-            path=f"{link_metadata.base_path}spaces/{link_metadata.space_key}/pages/{link_metadata.page_id}/{link_metadata.title}",
             path=os.path.join(link_metadata.base_path, "display", link_metadata.space_key, relative_path),
             params="",
             query="",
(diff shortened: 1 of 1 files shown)


md2conf/converter.py
```diff
@@ -301,6 +302,7 @@ def _transform_link(self, anchor: ET._Element) -> None:
             scheme="https",
             netloc=link_metadata.domain,
             path=f"{link_metadata.base_path}spaces/{link_metadata.space_key}/pages/{link_metadata.page_id}/{link_metadata.title}",
+            path=os.path.join(link_metadata.base_path, "display", link_metadata.space_key, relative_path),
```
AdamGagorik:
```suggestion
            path=os.path.join(link_metadata.base_path, "display", link_metadata.space_key, relative_path),
```

md2conf/converter.py
```diff
@@ -281,6 +281,7 @@ def _transform_link(self, anchor: ET._Element) -> None:
                 raise DocumentError(msg)
 
         relative_path = os.path.relpath(absolute_path, self.base_path)
+        relative_path = os.path.splitext(os.path.relpath(absolute_path, self.base_path))[0]
```
AdamGagorik:
```suggestion
        relative_path = os.path.splitext(os.path.relpath(absolute_path, self.base_path))[0]
```

hunyadi:
Personally, I would prefer the *PageId Format* (used in current `master`) as opposed to the *Page Title ("Pretty") Format* (introduced in this pull request). The *PageId Format* builds links based on a globally unique identifier, and is resilient to page title changes. This is why Markdown documents synchronized with *md2conf* are expected to have a *page ID* reference.

I am no Confluence expert but the [knowledge base page](https://confluence.atlassian.com/confkb/the-differences-between-various-url-formats-for-a-confluence-page-278692715.html) you are linking in the pull request description seems to relate to products that reached their end of life on February 15th 2024. If this alternative way of building URLs to other pages is available in officially supported current versions of Confluence, I don't mind adding this capability triggered by a configuration option.

PR #34, opened 2024-04-28
Add support for directory structures

### LINKED ISSUE #26: Is there any way to export a subfolder as a file?
Hello, first of all thanks for this great work.
Is there any way to export a subfolder as a file? 
When I export a subfolder that contains some .md files they are being exported to the root page id (the one defined by the argument: -r $page_id) so there´s no way to have child pages like the ones used in Confluence.
Any workaround for this?

Thank you!

ojacques:
- New: Fix #26: properly handle directory structure
- New: if page name already exists, create a page by appending invisible space in title
- New: use markdown title (# Title) as title for the page
- Fix: don't raise exception when not finding page in get_page_id_by_title
- Fix: Add confluence keys in markdown pages after frontmatter instead of always at the top (fix for mkdocs-material)
- Fix: If admonition type is not known, default to info instead of raising exception

### DIFF
--- a/md2conf/api.py
+++ b/md2conf/api.py
@@ -27,6 +27,25 @@ JsonType = Union[
     List["JsonType"],
 ]
 
+# Array of 16 Unicode characters which show as spaces, and workaround Confluence page name uniqueness
+SPACE_CHARACTERS = [
+    "\u00A0",
+    "\u2000",
+    "\u2001",
+    "\u2002",
+    "\u2003",
+    "\u2004",
+    "\u2005",
+    "\u2006",
+    "\u2007",
+    "\u2008",
+    "\u2009",
+    "\u200A",
+    "\u200B",
+    "\u202F",
+    "\u205F",
+    "\u3000",
+]
 
 def build_url(base_url: str, query: Optional[Dict[str, str]] = None) -> str:
     "Builds a URL with scheme, host, port, path and query string parameters."
@@ -122,6 +141,7 @@ class ConfluenceSession:
     domain: str
     base_path: str
     space_key: str
+    nb_duplications: int = 0
 
     def __init__(
         self, session: requests.Session, domain: str, base_path: str, space_key: str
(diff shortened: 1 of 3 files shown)


hunyadi (commented):
Thanks for your contribution, it's a great addition to the feature set of [md2conf](https://pypi.org/project/markdown-to-confluence/)! I am not sure I agree with using Unicode spaces to magically mutate page titles, which may adversely impact user experience as they want to distinguish two pages. I would also want to have a few integration tests in place (see the directories `sample` and `tests`) before the new features are pushed to `master` to avoid future regressions. However, I am absolutely OK to merge these changes to a `dev` branch where we can continue to collaborate.

ojacques:
Thanks @hunyadi for your kind words. I needed a Python project, and yours was the closest to providing the results I wanted. Great work!

Regarding the spaces, I have to admit it is a bit of a hack 😅. Now, Confluence does not allow duplicate page names in a given space, and this happens quite a bit in my case (multiple 'introduction', 'template' and other non-unique names for pages on a large documentation site). Not to mention that this documentation is also published in parallel to other folders in the same space, which I do not control at all. 

As an alternative to hidden spaces, I tried just numbers ('Introduction 1', 'Template 2'), but I found it... Not pretty. I'm curious to read other ideas you might have. 
Maybe we should have multiple options, giving the choice on the command line?

hunyadi:
A rather straightforward solution to the conflicting title issue could be to use a breadcrumb-style approach to building titles: use the parent page title as a prefix for the child page, e.g. assuming the parent page has the title *Quest of the Avatar*, and the child page has the title *Magic*, the synthesized unique title for the child page could be *Quest of the Avatar: Magic*. This could be applied recursively. My concern with the spaces is that while it solves the immediate issue with conflicting titles, it leads to cryptic behavior when the pages are referenced individually, and may make *md2conf* non-deterministic depending on the order the pages are published (i.e. they might be assigned a different space character).

PR #36, opened 2024-05-31
Extract test page id to a sep constant,  use Pathlib

vladistan:
Extracted the test page ID to a constant and put it at the top of the file.  This makes it easier to for other devs to work on this tool by modifying the code just in one place

Also,  switched to Path lib from os.path as it is a more modern alternative and it has been a part of Python standard lib for a while.  Probably should do it in the main code as well,  but need to get more familiar with it first

### DIFF
--- a/md2conf/__main__.py
+++ b/md2conf/__main__.py
@@ -26,7 +26,7 @@ class Arguments(argparse.Namespace):
     generated_by: Optional[str]
 
 
-def main():
+def main() -> None:
     parser = argparse.ArgumentParser()
     parser.prog = os.path.basename(os.path.dirname(__file__))
     parser.add_argument(
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -3,6 +3,7 @@ import os
 import os.path
 import shutil
 import unittest
+from pathlib import Path
 
 from md2conf.api import ConfluenceAPI, ConfluenceAttachment, ConfluencePage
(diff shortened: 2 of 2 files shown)


hunyadi:
That's definitely a simple but nevertheless great improvement, thanks for contributing!

As for `pathlib`, I would definitely prefer switching everywhere (or not at all), which would help avoid weird patterns like
```
synchronize(str(self.sample_dir / "example.md"))
```
in which there is no real need to convert to a `str` but instead we could pass a `pathlib.Path` directly. Using a single path formalism across the entire code base would also make it easier for readers to understand.

vladistan:
Yes,  this is just a first step into this direction.  I just wanted to get
a fell how receptive you'd be to PRs as some repo maintainers do not merge
PRs for years.

If you think you'd benefit from it I can go through the rest of the
codebase and patch it up

hunyadi:
If the pull request is focused and succinct (like yours), I am happy to merge. What I find more challenging is to merge pull requests that introduce incomplete features, or new features that pertain to specific use cases but impact the overall structure of the code. It's typically much harder to verify in these cases that the improvements don't break or adversely impact existing functionality.

vladistan:
Yes makes sense.

vladistan:
Closing in favor of https://github.com/hunyadi/md2conf/pull/37

PR #37, opened 2024-06-01
More democratic test fixes

vladistan:
Ok, now Path lib  is used throughout the code not only in the test,  so no need to do a lot of `str` conversions.

There are still couple of special cases where we do need to convert to `Path` and `str` I left comments for you

### DIFF
--- a/md2conf/application.py
+++ b/md2conf/application.py
@@ -1,5 +1,6 @@
 import logging
 import os.path
+from pathlib import Path
 from typing import Dict, Optional
 
 from .api import ConfluenceSession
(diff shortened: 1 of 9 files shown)


vladistan:
Ah,  also added the test case for processor

md2conf/application.py
```diff
@@ -25,40 +26,43 @@ def __init__(
...
-                    continue
+                # Reconstitute Path object back
+                docfile = (Path(root) / file_name).absolute()
 
-                absolute_path = os.path.join(os.path.abspath(root), file_name)
-                metadata = self._get_or_create_page(absolute_path)
+                # Skip non-markdown files
+                if docfile.suffix.lower() != ".md":
+                    continue
+                # check the file extension
```
hunyadi: Stray comment, should likely be removed.

md2conf/converter.py
```diff
@@ -270,9 +265,10 @@ def _transform_link(self, anchor: ET._Element) -> None:
         # convert the relative URL to absolute URL based on the base path value, then look up
         # the absolute path in the page metadata dictionary to discover the relative path
         # within Confluence that should be used
-        absolute_path = os.path.abspath(os.path.join(self.base_path, relative_url.path))
-        if not absolute_path.startswith(self.base_path):
-            msg = f"relative URL points to outside base path: {url}"
+        absolute_path = (self.base_path / relative_url.path).absolute()
+        # TODO: Looks bit suspicious, should look in path.resolve for better checking
```
hunyadi: This is only meant as a rather simple simple check to ensure we don't export paths that don't make sense in Confluence, e.g. `../../image.png` would be illegal in Confluence because there is no parent folder.

hunyadi (approved):

md2conf/application.py
vladistan: Yep..  fixed this

md2conf/converter.py
vladistan: I see...  so fixed this too

vladistan:
How does it look now?  GTG?

PR #40, opened 2024-06-15
Add test for converter

vladistan:
This one is to be merged after sibling PR with cleanup had been merged


vladistan:
This is extra unit test for your converter module

PR #42, opened 2024-08-16
Fix/use utf8 and ignore empty comments

zkygr:
Hello, i had 2 issues:

1) Emojis and umlaute were presented wrong in Confluence.
2) I could not update Confluence pages with attachements, when the attachments had empty comments.

### DIFF
--- a/md2conf/api.py
+++ b/md2conf/api.py
@@ -177,7 +177,8 @@ class ConfluenceSession:
         extensions = typing.cast(Dict[str, JsonType], result["extensions"])
         media_type = typing.cast(str, extensions["mediaType"])
         file_size = typing.cast(int, extensions["fileSize"])
-        comment = typing.cast(str, extensions["comment"])
+        comment = extensions.get("comment", "")
+        comment = typing.cast(str, comment)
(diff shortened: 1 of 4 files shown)


hunyadi (approved):

zkygr:
Hello and thank you for merging and the project :)
Are you planing to release a new version soon?
I would like to use this project at my work and install it accordingly through pypi.

Have a nice sunday

hunyadi:
The repository has another pull request open related to Mermaid diagrams, which I would also want to include in the upcoming version.

Until a new version is released, however, you can have `pip` install the package directly from the source repository:
```
pip install git+https://github.com/hunyadi/md2conf.git
```
We use this approach in several of our projects such that we don't have to wait for package releases. If you need to pin to a specific commit, you can add the commit hash at the end.

zkygr:
Thanks for the suggestion, but I don't think it will work because our internet access is restricted, and we can only access PyPI through a proxy, so we can't clone repositories directly. No worries though, I'll keep using my modified version for a bit longer.

hunyadi:
I see. I shall wait a few days on the other (open) pull request and if I don't get feedback, I will issue a version `0.1.13`.

Issue #45, opened 2024-08-28
Feature Request: Support for `<details><summary>` block


zender-vivodyne:
# Usecase

We are using markdown in Github to support collapsible sections. See [here](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections). Would like that to be converted into a Confluence `Expand Macro`

<img width="811" alt="image" src="https://github.com/user-attachments/assets/90b9d1b5-7478-4818-8bc8-27a346312387">

# Issues

Noticed today that the `markdown` python library doesn't seemingly parse things correctly when inside the `details/summary` blocks. For example if you nest a markdown list inside `details` it doesn't run it into a `<ul><li>`. Unsure if this issue is 2 things or not. 

1. PR into markdown to handle the `details/summary` blocks better so that html is returned correctly
2. PR here to turn any `details/summary` blocks into Confluence `Expand Macros`

Likely they are not dependent but would likely want both for to have correctly rendered data.

zendern:
On issue 1. above i did open this on the markdown library side as well. https://github.com/Python-Markdown/markdown/issues/1478

zendern:
> On issue 1. above i did open this on the markdown library side as well. [Python-Markdown/markdown#1478](https://github.com/Python-Markdown/markdown/issues/1478)

Looks like there is a plugin available to support this. I'll look at PR'ing this change should be simple enough. https://github.com/Python-Markdown/markdown/issues/1478#issuecomment-2318383029. That at least gets the first part of this to work. And then just need to figure out the confluence conversion part.

hunyadi:
This has been implemented on the `master` branch: *Add support for collapsed sections with expand macro*.

zendern:
> This has been implemented on the `master` branch: *Add support for collapsed sections with expand macro*.

@hunyadi thanks for turning around this feature so quickly. I did some testing on my docs and it worked mostly. I just opened #49 to fix one issue I found. Would love to see a new release soon. Let me know if I can help out in any other ways.

PR #46, opened 2024-08-30
Markdown in html

zendern:
Deals with part 1 of #45. This will at least allow users to use the `<details><summary>` blocks and have their content be rendered correctly. 

This ends up being an opt-in feature since the way `md-in-html` works is to require you to add an attribute to the html to tell it to actually process the contents and how to do so.


sample/markdown-in-html.md
```diff
@@ -0,0 +1,20 @@
+<!-- confluence-page-id: 85668266616 -->
+## Inside Details/Summary block
+
+<!-- If you add the attribute of markdown it'll be converted to html -->
+<details markdown="1">
+<summary><b>My summary</b></summary>
```
zendern, line 6: here you can see users will have to opt in to use it. And the div below just keeps all the pipe symbols around vs turning it into a table.

README.md
```diff
@@ -20,6 +20,7 @@ This Python package
 * Image references (uploaded as Confluence page attachments)
 * [Table of Contents](https://docs.gitlab.com/ee/user/markdown.html#table-of-contents)
 * [Admonitions](https://python-markdown.github.io/extensions/admonition/) (converted into *info*, *tip*, *note* and *warning* Confluence panels)
+* [Markdown in HTML](https://python-markdown.github.io/extensions/md_in_html/)
```
zendern, line 23: Unsure if we should add more docs around this or just a link to the plugin like you have done for admonition is good enough 🤷

hunyadi (approved):

PR #51, opened 2024-09-08
Adding CI/CD

hemstreet:
Adding CI/CD to build a docker image and publish to dockerhub as well as build a tarball / wheel for pypi. These actions will be triggered when new release is published.

You will need to add three secrets to the project to publish to both sources.

- **DOCKER_USERNAME** ( this PR assumes your account is hunyadi )
- **DOCKER_PASSWORD** ( Docker API key )
- **PYPI_ID_TOKEN**

This will also package up the latest code with the most recent [Mermaid and Kroki  ](https://github.com/hunyadi/md2conf/commit/a1e29d3bd38285aad520178000313815ab33b7a7) updates. 

The docker image will operate the exact same way as the python module
`docker run --rm hunyadi/md2conf -d instructure.atlassian.net ./`

These will create releases like:
[Docker hub](https://hub.docker.com/r/hemstreet/md2conf) and [PyPi](https://pypi.org/project/hemstreet-markdown-to-confluence/). Note that I’ll delete these if this gets merged. Was just for testing / making sure it was aligned with how it’s getting published today.


README.md
```diff
@@ -149,3 +149,30 @@ optional arguments:
   --ignore-invalid-url  Emit a warning but otherwise ignore relative URLs that point to ill-specified locations.
   --local               Write XHTML-based Confluence Storage Format files locally without invoking Confluence API.
 ```
+
+### Using the docker container
+
+You can run the dokcer container via docker run or via dockerfile. Either can accept the environment variables or argument similar to the python options.
```
zendern:
```suggestion
You can run the docker container via docker run or via dockerfile. Either can accept the environment variables or argument similar to the python options. 
```

README.md
```diff
@@ -149,3 +149,30 @@ optional arguments:
   --ignore-invalid-url  Emit a warning but otherwise ignore relative URLs that point to ill-specified locations.
   --local               Write XHTML-based Confluence Storage Format files locally without invoking Confluence API.
 ```
+
+### Using the docker container
+
+You can run the dokcer container via docker run or via dockerfile. Either can accept the environment variables or argument similar to the python options. 
+The final argument `./` is mdpath.
+
+* `docker run --rm --name md2conf hemstreet/md2conf -d instructure.atlassian.net -u levente.hunyadi@instructure.com -a 0123456789abcdef -s DAP ./`
```
zendern: Assume we will want to make this to match the github repo? so `hunyadi/md2conf`?

Dockerfile_dist
```diff
@@ -0,0 +1,25 @@
+ARG PYTHON_VERSION=3.11
```
zendern: Might as well make this 3.12. The library supports it and no reason to not be on the latest.

README.md
hemstreet: Good catch, absolutely right.

Dockerfile_dist
hemstreet: updated

.github/workflows/publish-python.yml
```diff
@@ -0,0 +1,29 @@
...
+    name: Publish release to PyPI
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write  # IMPORTANT: mandatory for sigstore
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
```
hunyadi: While it's likely less relevant in this case, we might want to use the earliest Python version that the package officially supports to bulid the distribution. In `setup.cfg`, we have:
```
python_requires = >=3.9
```

md2conf/__init__.py
```diff
@@ -5,7 +5,7 @@
 Confluence API endpoints to upload images and content.
 """
 
-__version__ = "0.1.13"
+__version__ = "0.1.14"
```
hunyadi: Please don't increment the version number yet. Several new features have been added recently. This is likely to be part of `0.2.0`.

hunyadi (approved):

.github/workflows/publish-python.yml
hemstreet: this has been changed to 3.9, I picked 3.11 because thats what was in the docker container originally.

https://github.com/hunyadi/md2conf/blob/master/Dockerfile#L1

.github/workflows/publish-python.yml
```diff
@@ -0,0 +1,29 @@
...
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.9"
+      - name: Install dependencies
+        run: |
+          python3 -m pip install --upgrade pip
+          pip3 install wheel
+      - name: Build package
+        run: |
+          python3 setup.py sdist bdist_wheel
```
hunyadi: I am afraid we are somewhat inconsistent how we build the package. Specifically, we use `setup.py` here but `pip wheel` in the Docker container. For other projects, we typically use [build](https://pypi.org/project/build/). Unless there is a specific reason to keep these as-is, I would prefer we use `build` both in the *Publish to PyPI* and the Docker distribution actions.

hemstreet: I made these both consistent now. There was not a specific reason for the inconsistency. Just coding on two different days with tunnel vision. Nice catch.

hunyadi (approved):

PR #63, opened 2024-10-02
Improve handling of local anchor links

erauner12:


md2conf/__init__.py
```diff
@@ -5,7 +5,7 @@
 Confluence API endpoints to upload images and content.
 """
 
-__version__ = "0.2.2"
+__version__ = "0.2.3"
```
hunyadi, line 8: Please don't increment the version number because other changes might get on the same release train.

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [metadata]
-name = markdown-to-confluence
+name = md2conf
```
hunyadi, line 2: The name of the package in PyPI is indeed `markdown-to-confluence`. Unfortunately, when I was about to publish *md2conf* to PyPI, the name `md2conf` was already taken.

md2conf/converter.py
```diff
@@ -317,6 +317,12 @@ def _transform_link(self, anchor: ET._Element) -> None:
         LOGGER.debug(f"found link {url} relative to {self.path}")
         relative_url: ParseResult = urlparse(url)
 
+        if relative_url.fragment and not relative_url.path:
+            # This is a local anchor link
+            LOGGER.debug(f"found local anchor URL: {url}")
+            anchor.attrib["href"] = f"ac:link ac:anchor='{relative_url.fragment}'"
+            return
```
hunyadi, line 324: This seems like magic. Why would we want to inject a series of qualified attribute name and value pairs into the `href` attribute of the anchor element? If this works, it appears that we are exploiting a bug in `lxml`. I assume that we might want to construct a new element (either an anchor or a structured macro) and populate the fields based on what attributes the original anchor element had. There are a few examples in the source code how to do this.

hunyadi (changes requested):
Thanks for the pull request! I have added some comments in the hope that the approach here better aligns with how similar transformations are treated today, such that future maintainers see a more uniform code base.

hunyadi:
Local references within a Markdown document (i.e. URLs that consist of a *fragment* part only) are now expanded into Confluence advanced links:
```
<ac:link ac:anchor="anchor">
  <ac:link-body>
    ...
  </ac:link-body>
</ac:link>
```
This is activated with `--heading-anchors`. Links must use standard Markdown conversion rules to reference section headings (all lowercase).

A more robust solution than proposed in this pull request has been pushed to `master` with the commit text *Expand references within document to Confluence advanced links*.

PR #71, opened 2024-11-15
Calls to Path.absolute() do not resolve relative paths

Cau5tik:
Hey folks. This is my first PR to someone else's public project. I've done my best to be thorough in my writing, but I'm sure this is bad for some reason I'm not aware of. Please let me know if there's anything I'm missing or anything that should be better. Thanks!

Given a sample directory structure:

```
/parent
 |
 |_> /child
       |_> document.md
 |_> README.md
```

Where `/parent/child/document.md` contains a reference to `/parent/README.md`:
```
[See README](../README.md)
```

`_transform_link` will throw a DocumentError:
```
2024-11-14 15:44:37,959 - INFO - _synchronize_page [88] - Synchronizing page: /Users/sliddell/Documents/parent/child.md
Traceback (most recent call last):
  File "/Users/sliddell/.pyenv/versions/3.9.17/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Users/sliddell/.pyenv/versions/3.9.17/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/sliddell/Documents/scratch/md2conf/lib/python3.9/site-packages/md2conf/__main__.py", line 222, in <module>
    main()
  File "/Users/sliddell/Documents/scratch/md2conf/lib/python3.9/site-packages/md2conf/__main__.py", line 204, in main
    Application(
  File "/Users/sliddell/Documents/scratch/md2conf/lib/python3.9/site-packages/md2conf/application.py", line 49, in synchronize
    self.synchronize_directory(path)
  File "/Users/sliddell/Documents/scratch/md2conf/lib/python3.9/site-packages/md2conf/application.py", line 79, in synchronize_directory
    self._synchronize_page(page_path, page_metadata)
  File "/Users/sliddell/Documents/scratch/md2conf/lib/python3.9/site-packages/md2conf/application.py", line 89, in _synchronize_page
    document = ConfluenceDocument(page_path, self.options, page_metadata)
  File "/Users/sliddell/Documents/scratch/md2conf/lib/python3.9/site-packages/md2conf/converter.py", line 1003, in __init__
    converter.visit(self.root)
  File "/Users/sliddell/Documents/scratch/md2conf/lib/python3.9/site-packages/md2conf/converter.py", line 263, in visit
    self.visit(source)
  File "/Users/sliddell/Documents/scratch/md2conf/lib/python3.9/site-packages/md2conf/converter.py", line 259, in visit
    target = self.transform(source)
  File "/Users/sliddell/Documents/scratch/md2conf/lib/python3.9/site-packages/md2conf/converter.py", line 830, in transform
    return self._transform_link(child)
  File "/Users/sliddell/Documents/scratch/md2conf/lib/python3.9/site-packages/md2conf/converter.py", line 406, in _transform_link
    raise DocumentError(msg)
md2conf.converter.DocumentError: unable to find matching page for URL: ../README.md
```

Setting a breakpoint at [line 399 in converter.py](https://github.com/hunyadi/md2conf/blob/master/md2conf/converter.py#L399) shows that the PosixPath of `absolute_path` is _not_ resolved as per the comments at [line 383](https://github.com/hunyadi/md2conf/blob/master/md2conf/converter.py#L383)

```
-> if link_metadata is None:
(Pdb) self.page_metadata
{PosixPath('/Users/sliddell/Documents/oncall-runbooks/parent/README.md'): ConfluencePageMetadata(domain='<not relevant>', base_path='/wiki/', page_id='1222148237', space_key='ENG', title='sliddelltest'), PosixPath('/Users/sliddell/Documents/oncall-runbooks/parent/child/document.md'): ConfluencePageMetadata(domain='<not_relevant>', base_path='/wiki/', page_id='1222148247', space_key='ENG', title='Document')}
(Pdb) absolute_path
PosixPath('/Users/sliddell/Documents/oncall-runbooks/parent/child/../README.md')
```

The [PathLib docs for Path.absolute](https://docs.python.org/3/library/pathlib.html#pathlib.Path.absolute) explicitly says that `absolute()` does _not_ normalize relative paths. Looking at [Path.resolve below it](https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve), this function _does_ resolve relative paths. It also calls out that this is the _only_ function that will eliminate `..` components.

This means that if the value for `absolute_path` in `_transform_link` is a relative path, it will never match a key in the `self.page_metadata` dictionary. We need to resolve the relative path into an absolute path when we want to compare it to the keys of `self.page_metadata`.

Changing [line 386](https://github.com/hunyadi/md2conf/blob/master/md2conf/converter.py#L386) to `resolve()` rather than `absolute()` doesn't fix the issue, as we're referring to a markdown document in another directory. This will raise a different DocumentError that the relative URL is outside the base path. While this is true, this doesn't actually break anything. Resolving the relative path when comparing it to the `page_metadata` dictionary results in the correct link being created to the markdown document in the parent folder.

If this is better fixed at line 386 when we expect links to resolve, then it would be great to better understand why you don't want to traverse directories with relative paths. My use case involves multiple teams creating documentation in a single repository. There are enough teams and enough documents that not everything will live in a single folder, but we would still like to make references between documents to keep them DRY.

Please let me know if there's anything else I can provide. Thanks for your time and efforts!

### DIFF
--- a/md2conf/converter.py
+++ b/md2conf/converter.py
@@ -395,7 +395,7 @@ class ConfluenceStorageFormatConverter(NodeVisitor):
 
         relative_path = os.path.relpath(absolute_path, self.base_path)
 
-        link_metadata = self.page_metadata.get(absolute_path)
+        link_metadata = self.page_metadata.get(absolute_path.resolve())
         if link_metadata is None:
             msg = f"unable to find matching page for URL: {url}"
             if self.options.ignore_invalid_url:
--- a/md2conf/converter.py
+++ b/md2conf/converter.py
@@ -301,7 +301,8 @@ class ConfluenceStorageFormatConverter(NodeVisitor):
 
     options: ConfluenceConverterOptions
     path: Path
-    base_path: Path
+    base_dir: Path
+    root_dir: Path
     links: List[str]
     images: List[str]
     embedded_images: Dict[str, bytes]
@@ -311,12 +312,14 @@ class ConfluenceStorageFormatConverter(NodeVisitor):
         self,
         options: ConfluenceConverterOptions,
         path: Path,
+        root_dir: Path,
         page_metadata: Dict[Path, ConfluencePageMetadata],
     ) -> None:
         super().__init__()
         self.options = options
         self.path = path
-        self.base_path = path.parent
+        self.base_dir = path.parent
+        self.root_dir = root_dir
         self.links = []
         self.images = []
         self.embedded_images = {}
@@ -383,9 +386,9 @@ class ConfluenceStorageFormatConverter(NodeVisitor):
         # convert the relative URL to absolute URL based on the base path value, then look up
         # the absolute path in the page metadata dictionary to discover the relative path
         # within Confluence that should be used
-        absolute_path = (self.base_path / relative_url.path).absolute()
-        if not str(absolute_path).startswith(str(self.base_path)):
-            msg = f"relative URL {url} points to outside base path: {self.base_path}"
+        absolute_path = (self.base_dir / relative_url.path).resolve(True)
+        if not str(absolute_path).startswith(str(self.root_dir)):
+            msg = f"relative URL {url} points to outside root path: {self.root_dir}"
             if self.options.ignore_invalid_url:
                 LOGGER.warning(msg)
                 anchor.attrib.pop("href")
@@ -393,9 +396,7 @@ class ConfluenceStorageFormatConverter(NodeVisitor):
             else:
                 raise DocumentError(msg)
 
-        relative_path = os.path.relpath(absolute_path, self.base_path)
-
-        link_metadata = self.page_metadata.get(absolute_path.resolve())
+        link_metadata = self.page_metadata.get(absolute_path)
         if link_metadata is None:
             msg = f"unable to find matching page for URL: {url}"
             if self.options.ignore_invalid_url:
@@ -405,6 +406,7 @@ class ConfluenceStorageFormatConverter(NodeVisitor):
             else:
                 raise DocumentError(msg)
 
+        relative_path = os.path.relpath(absolute_path, self.base_dir)
         LOGGER.debug(
             "found link to page %s with metadata: %s", relative_path, link_metadata
         )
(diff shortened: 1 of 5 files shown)


md2conf/converter.py
```diff
@@ -395,7 +395,7 @@ def _transform_link(self, anchor: ET._Element) -> Optional[ET._Element]:
 
         relative_path = os.path.relpath(absolute_path, self.base_path)
 
-        link_metadata = self.page_metadata.get(absolute_path)
+        link_metadata = self.page_metadata.get(absolute_path.resolve())
```
hunyadi: We do invoke `pathlib.Path.resolve()` in `Application.synchronize()` but I agree that this is not sufficient given the use case you are describing. When concatenating relative paths starting with `..` to the document base path as we are transforming links in the Markdown document, we can end up with paths that are no longer canonical, and the code implicitly assumes that all paths are canonical, hence the error you have encountered.

Checking containment in the base path (few lines above) is a legitimate check, many applications do this for safety. However, when exporting a directory hierarchy, this base path should be the topmost directory we are exporting, rather than the document's location (as it is now), such that you can cross-reference documents in the directory hierarchy the way you describe.

I will look into possible solutions.

hunyadi:
Changes have been pushed to the branch `master` that set both a `base_dir` and a `root_dir` in the initializer of `ConfluenceStorageFormatConverter`. As before, `base_dir` is used for resolving relative links such as `path/to/other.md` in the Markdown document. In contrast, `root_dir` is used for checking whether the linked document is within the directory hierarchy being published. The combination of these member variables should allow you to use relative links in Markdown that "go up" (e.g. `../other.md`) so long as they don't go beyond the directory you originally passed to *md2conf* as a command-line argument.

Feel free to let me know if these changes have addressed the issue you reported.

Last but not least, thank you for reporting the issue, and providing an in-depth description, which has helped tremendously to track down the root cause.

hunyadi (approved):

Cau5tik:
Works great. Relative URLs to parent directories are working very well.
Thank you for such a fast response!

Issue #83, opened 2025-02-19
Implement a preserve-structure option


ruispereira:
The goal of this feature is to detect folders and put all files and sub-folders below while sending the markdown to confluence.
This way, all detected folders that contains files/sub-folders will have a parent page.

hunyadi:
Do I understand it correctly that your proposal would have *md2conf* create a (possibly virtual) `index.md` or `README.md` in each subdirectory, and synchronize the directory hierarchy with Confluence as if these (virtual) files were originally present? In other words, each directory entry in the filesystem would correspond to a blank page in Confluence (or a page possibly with a table of contents), and *md2conf* would assign all Markdown files in the directory to this blank page as children.

ruispereira:
Yes, imagine that you have a structure like:
- file1.md
- folder1
- -file2.md
- -file3.md

In this example, the folder1 will have a corresponding blank page called folder1 with all the files below.

ruispereira:
I can't push my branch, can you give me permissions to push my feature branch please?
--
git push --set-upstream origin preserve-structure-feature
ERROR: Permission to hunyadi/md2conf.git denied to ruispereira.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

hunyadi:
Today, you accomplish this with `index.md` or `README.md`. Assume you have the following directory listing:

```
.
├── file1.md
└── folder1
    ├── index.md
    ├── file2.md
    └── file3.md
```

This will produce the following page hierarchy in Confluence:

```
root
├── Page for file1.md
└── Page for index.md
    ├── Page for file2.md
    └── Page for file3.md
```

We need a separate file such as `index.md` or `README.md` to store the binding between the Markdown file and Confluence page ID. Normally, this information is saved at the beginning of the Markdown file as a Markdown comment.

However, I agree that it would be useful for *md2conf* to create these `index.md` or `README.md` files automatically with some default content.

If you want to contribute code, you should first fork the repository. You have full ownership of the fork, and can push any changes you want. If you then want to merge your changes back to the mainline branch (in the official *md2conf* repository), you should open a pull request.

PR #84, opened 2025-02-19
Implement preserve-structure feature

ruispereira:
Implement preserve-structure feature like described here:
https://github.com/hunyadi/md2conf/issues/83

### DIFF
--- a/md2conf/application.py
+++ b/md2conf/application.py
@@ -142,8 +142,9 @@ class Application:
             convert_to_filename = local_dir.name + ".md"
             parent_doc = Path(local_dir) / convert_to_filename
             is_folder = True
-            with open(parent_doc, "w") as f:
-                pass
+            if not os.path.exists(parent_doc):
+                with open(parent_doc, "w") as f:
+                    pass
 
         if parent_doc is not None:
             if parent_doc in files:
--- a/md2conf/application.py
+++ b/md2conf/application.py
@@ -23,7 +23,6 @@ from .converter import (
     read_qualified_id,
 )
 from .matcher import Matcher, MatcherOptions
-from pathlib import Path
 
 LOGGER = logging.getLogger(__name__)
 
@@ -87,12 +86,9 @@ class Application:
         self._index_directory(local_dir, root_id, page_metadata)
         LOGGER.info("Indexed %d page(s)", len(page_metadata))
 
-        # filter ignored pages from map
-        filtered_page_metadata = {key: value for key, value in page_metadata.items() if not value.ignore}
-
         # Step 2: convert each page
-        for page_path in filtered_page_metadata.keys():
-            self._synchronize_page(page_path, root_dir, filtered_page_metadata)
+        for page_path in page_metadata.keys():
+            self._synchronize_page(page_path, root_dir, page_metadata)
 
     def _synchronize_page(
         self,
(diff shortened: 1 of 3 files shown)


md2conf/application.py
```diff
@@ -131,12 +135,25 @@ def _index_directory(
             parent_doc = Path(local_dir) / "index.md"
         elif (Path(local_dir) / "README.md") in files:
             parent_doc = Path(local_dir) / "README.md"
+        
+        is_folder = False
+        if self.options.preserve_structure:
+            # current dirname converted to filename
+            convert_to_filename = local_dir.name + ".md"
```
hunyadi: Is there a specific reason why we choose to make this file have the same name as the parent directory instead of `index.md` or `README.md`? This seems to introduce an inconsistency between the cases when the command-line option `preserve-structure` is used vs. not used; and we need to update the documentation to explain the two use cases and how they are different.

On the other hand, if we simply created a new `index.md` or `README.md` whenever a directory has none of these files, the two modes would behave consistently. No extra documentation updates are needed.

If we assign `parent_doc` but `parent_doc` is not present in the file system, we could create a new file with `open` (unless the current directory has no subdirectories to recurse into, and thus cannot have child pages in Confluence). This would also eliminate the need for `ignore` in `ConfluencePageMetadata`, as well as the local variable `is_folder`.

The reason why the names `index.md` and `README.md` were originally chosen was that GitHub automatically displays one of these files if it finds them in a directory opened when browsing a project repository.

md2conf/application.py
```diff
@@ -131,12 +135,25 @@ def _index_directory(
...
+            parent_doc = Path(local_dir) / convert_to_filename
+            is_folder = True
+            with open(parent_doc, "w") as f:
+                pass
```
hunyadi: Be careful as this may overwrite an existing file of the same name.

md2conf/converter.py
```diff
@@ -246,7 +246,7 @@ class ConfluencePageMetadata:
     page_id: str
     space_key: str
     title: str
-
+    ignore: Optional[bool] = False
```
hunyadi: By structuring `_index_directory` a little differently, I think we can avoid having to add this field, which is otherwise not used elsewhere.

hunyadi (commented):
Thank you for making and sharing these changes. I have added some comments to simplify the code further, and ensure we have a consistent user experience. Otherwise, I am happy to merge.

md2conf/converter.py
ruispereira: Actually, this field is used in Application.py, to mark the virtual files with the name of the directory to be ignore, as we need to avoid synchronising pages like this. 
Check lines 88 and 156

md2conf/application.py
ruispereira: nice catch, will change that

md2conf/application.py
ruispereira: The idea is to have the same structure that i already have in my repo for example. I don't want to have to create a new page, or have a page named index nor a readme as a parent page. I want a custom name, something meaningful. That's the idea here. And using the folder name, i don't need to create more logic, as what i want is the same as i had left in my repo. One of the points of improvement is trying to known if the file is already created (with content or not), or it doesn't exist (like i have currently) and needs to be created on-the-fly. Makes sense?

hunyadi: I totally understand the reasoning for the alternative name. While `index.md` and `README.md` are used by GitHub, they are still arbitrary choices. Having a file whose name matches the parent directory name as the index file to assume parent role makes perfect sense, we can definitely add that as another option to check for.

What eludes me is why we need the complexity with `is_folder` and `ignore`. We could just add an extra check for `DIRNAME.md` at the beginning (like we have for `index.md`), potentially create the file if it does not exist (assuming `preserve_structure` is enabled), and proceed as before. This gives fewer lines of code for future maintainers to go through.

I definitely don't want to push work onto you, I am happy to make these changes, and merge into `master`.

hunyadi: I have pushed [some code](https://github.com/hunyadi/md2conf/blob/dev/md2conf/application.py) to the branch `dev` to show what I mean.

ruispereira: Hi @hunyadi, i check the code you pushed. 
Checking the code, your suggestion is that we have to create the dir_name.md manually, right?

elif (Path(local_dir) / f"{local_dir.name}.md") in files:
            parent_doc = Path(local_dir) / f"{local_dir.name}.md"

ruispereira: My suggestion is, if the file not exists, but the user wants to keep the hierarchy, create the local_dir_name.md on-the-fly, like i did.

ruispereira:
```
parent_doc: Optional[Path] = None
if (Path(local_dir) / "index.md") in files:
    parent_doc = Path(local_dir) / "index.md"
elif (Path(local_dir) / "README.md") in files:
    parent_doc = Path(local_dir) / "README.md"
elif (Path(local_dir) / f"{local_dir.name}.md") in files:
    parent_doc = Path(local_dir) / f"{local_dir.name}.md"

if parent_doc is None and self.options.preserve_structure:
    parent_doc = Path(local_dir) / f"{local_dir.name}.md"

    # create a blank page in Confluence for the directory entry
    if not os.path.exists(parent_doc):
        with open(parent_doc, "w"):
            pass
```

Like this, what do you think? Anyway, for me, i prefer to have to enable the flag and let the lib do the job for me.

hunyadi: We can make the name configurable, e.g. let the user decide if they want `index.md`, `README.md` or a file based on the parent directory name if keep hierarchy is enabled and there is no index page to assume parent node role.

PR #89, opened 2025-02-21
feat(mermaid.py): we need to be able to leverage the is_docker functionality with mmdc installed globally

braunreyes:
Per the issue described here for Ubuntu 23+

https://github.com/mermaid-js/mermaid-cli/blob/master/docs/linux-sandbox-issue.md

We need to add in some puppeteer configs if we want to use this in docker environment like github action runners

The logic already exists in code, but returns an mmdc executable path that will not work in gha.

This adds an envar that can be used to get the is_docker goodness but still use proper mmdc executable.

### DIFF
--- a/md2conf/mermaid.py
+++ b/md2conf/mermaid.py
@@ -29,7 +29,12 @@ def get_mmdc() -> str:
     "Path to the Mermaid diagram converter."
 
     if is_docker():
-        return "/home/md2conf/node_modules/.bin/mmdc"
+        # used to workaround known issue:
+        # https://github.com/mermaid-js/mermaid-cli/blob/master/docs/linux-sandbox-issue.md
+        if os.environ.get("MD2CONF_USE_ROOT_MMDC", "false") == "true":
+            return "mmdc"
+        else:
+            return "/home/md2conf/node_modules/.bin/mmdc"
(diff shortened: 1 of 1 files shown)


md2conf/mermaid.py
```diff
@@ -29,7 +29,12 @@ def get_mmdc() -> str:
     "Path to the Mermaid diagram converter."
 
     if is_docker():
-        return "/home/md2conf/node_modules/.bin/mmdc"
+        # used to workaround known issue:
+        # https://github.com/mermaid-js/mermaid-cli/blob/master/docs/linux-sandbox-issue.md
+        if os.environ.get("MD2CONF_USE_ROOT_MMDC", "false") == "true":
+            return "mmdc"
+        else:
+            return "/home/md2conf/node_modules/.bin/mmdc"
```
hunyadi: Originally, I had to resort to using the full path in a Docker environment because simply invoking `mmdc` would not work. However, we could check whether the full path points to a valid file, and if not, fall back to `mmdc`. This might be a more flexible approach than adding a (currently undocumented) environment variable `MD2CONF_USE_ROOT_MMDC`.

hunyadi (approved):

PR #92, opened 2025-03-05
Remove type=global from /spaces api call

mikecsmith:
**Context**

Passing the `type=global` query param in the `spaces` API call prevents `md2conf` from being used with `knowledge_base` spaces.

**Note**

I did consider allowing the user to pass in the type via a `--space-type` flag with a default to `global` but my Python isn't fantastic. That feels like it would be a significant improvement over this PR but may not be necessary.

**Contributing**

I've read `CONTRIBUTING.md` and ran `check.sh` prior to raising the PR.

### DIFF
--- a/md2conf/api.py
+++ b/md2conf/api.py
@@ -194,7 +194,7 @@ class ConfluenceSession:
             payload = self._invoke(
                 ConfluenceVersion.VERSION_2,
                 "/spaces",
-                {"ids": id, "type": "global", "status": "current"},
+                {"ids": id, "status": "current"},
             )
             payload = typing.cast(dict[str, JsonType], payload)
             results = typing.cast(list[JsonType], payload["results"])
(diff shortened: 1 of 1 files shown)


md2conf/api.py
```diff
@@ -194,7 +194,7 @@ def space_id_to_key(self, id: str) -> str:
             payload = self._invoke(
                 ConfluenceVersion.VERSION_2,
                 "/spaces",
-                {"ids": id, "type": "global", "status": "current"},
+                {"ids": id, "status": "current"},
```
hunyadi, line 197: Looking at Confluence REST API v2 documentation for [spaces](https://developer.atlassian.com/cloud/confluence/rest/v2/api-group-space/#api-spaces-get), `type` doesn't appear to be a required parameter. Originally, functions *space_id_to_key* and *space_key_to_id* were eager, and they would fetch **all** spaces in Confluence, which is why I added the filter to `global` to ensure we have fewer results. However, I have recently switched to lazy evaluation, with the space key traded to space ID (or vice versa) when necessary, e.g. a page is about to be created, and the space key in the Markdown file has to be exchanged to a space ID. This means that the query response has at most a single result, and we no longer need the filter to `global`.

mikecsmith: Thanks for the context and much appreciated on the fast review/approval. I wasn't sure if the `type: "global"` had a specific use case but this helps to clear that up.

Brilliant project and thanks for open sourcing it!

mikecsmith:
Hi @hunyadi - is there any chance you could trigger a release with the updated code? I'm currently still deploying off my fork and would love to move over to the built docker image.

Thanks!

PR #94, opened 2025-03-28
Add parent_id and parent_type in ConfluencePage data structure

deleze:
Add `parent_id` and `parent_type` in ConfluencePage to avoid to make additional requests to get this information.

### DIFF
--- a/md2conf/api.py
+++ b/md2conf/api.py
@@ -40,6 +40,17 @@ class ConfluenceVersion(enum.Enum):
     VERSION_2 = "api/v2"
 
 
+class ConfluencePageParentContentType(enum.Enum):
+    """
+    Content types that can be a parent to a Confluence page
+    """
+    PAGE = "page"
+    WHITEBOARD = "whiteboard"
+    DATABASE = "database"
+    EMBED = "embed"
+    FOLDER = "folder"
+
+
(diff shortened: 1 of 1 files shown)


md2conf/api.py
```diff
@@ -71,6 +71,8 @@ class ConfluenceAttachment:
 class ConfluencePage:
     id: str
     space_id: str
+    parent_id: str
+    parent_type: str
```
hunyadi: `parent_type` seems more like an enumeration type of string values rather than a plain `str`. Specifically, [Confluence REST API documentation](https://developer.atlassian.com/cloud/confluence/rest/v2/api-group-page/#api-pages-id-get) writes:

> `parentType`: **ParentContentType**
> Content type of the parent, or null if there is no parent.
> Valid values: `page`, `whiteboard`, `database`, `embed`, `folder`

It might make sense to introduce a new *enum* type called `ParentContentType` with the values `page`, `whiteboard`, `database`, `embed` and `folder`, which would help those downstream users that work with the data-class `ConfluencePage`.

hunyadi (approved):

md2conf/api.py
deleze: Yes good idea! 
Done

hunyadi (approved):

PR #99, opened 2025-04-23
if front matter does not contain atitle, use first heading as title

gkoulin:
Extracts heading title from a Markdown document. Match first line starting with `#`.


md2conf/application.py
```diff
@@ -176,7 +171,11 @@ def _get_or_create_page(
             document = f.read()
 
         qualified_id, document = extract_qualified_id(document)
-        frontmatter_title, _ = extract_frontmatter_title(document)
+
+        if title is None:
+            title, _ = extract_frontmatter_title(document)
+        if title is None:
+            title, _ = extract_first_heading_title(document)
```
hunyadi: There is a corner case in which the front-matter exists but has no `title` attribute. In this case, the regular expression in the function `extract_first_heading_title` might scan into the front-matter, and accidentally match against something that is not Markdown (e.g. a `#` character in a quoted string). A more robust approach would be to use the second return value of `extract_frontmatter_title`, which is the text without the front-matter.

md2conf/converter.py
```diff
@@ -954,6 +954,11 @@ def extract_frontmatter_title(text: str) -> tuple[Optional[str], str]:
 
     return title, text
 
+def extract_first_heading_title(text: str) -> tuple[Optional[str], str]:
+    "Extracts heading title from a Markdown document. Match first line starting with `#`."
+
+    return extract_value(r'(?ms)^#\s+(.*?)$', text)
```
hunyadi: Personally, I would prefer setting modifiers `MULTILINE` and `DOTALL` explicitly, rather than embedding these instructions into the regular expression as `(?ms)`, which I find less readable. This would, however, entail modifying the signature of `extract_value`. I can take this as an exercise.

md2conf/converter.py
```diff
@@ -1035,6 +1040,10 @@ def __init__(
         # extract frontmatter
         self.title, text = extract_frontmatter_title(text)
 
+        # extract first heading
+        if self.title is None:
+            self.title, text = extract_first_heading_title(text)
```
hunyadi, line 1045: By overwriting the value assigned to the variable `text` with the second return value of `extract_first_heading_title` we are discarding the first heading in the document, which is likely not what we want. Instead, we should use the original document (with front-matter removed by `extract_frontmatter_title`).

md2conf/converter.py
giorgek-rfpro: I would agree with this. I followed the convention already taken, e.g. [converter.py](https://github.com/gkoulin/md2conf/blob/7dc7b951fa3032bc265411243b7b9de1389ead55/md2conf/converter.py#L941)

md2conf/converter.py
giorgek-rfpro: I would argue that you do want to remove the title. Otherwise, the page would have repeated text, firstly in the title of the page and secondly as a heading in the content.

md2conf/converter.py
hunyadi: Do we need to pass `re.DOTALL` (a.k.a. the modifier `s`)? The Markdown heading text cannot have a linefeed, so `.` in the regular expression doesn't need to capture `\n`. The expression `(.*?)` is ungreedy, so it will yield the same result but if we can simplify and eliminate an extra modifier, why not? `(?m)^#\s+(.*)$` seems a more straightforward alternative.

(We definitely need `re.MULTILINE`.)

md2conf/converter.py
hunyadi: That's true. On the other hand, some of our documents have an introductory text, then a Markdown heading, and then the rest of the Markdown document. (Furthermore, we have the preamble block, which is shown after the Confluence wiki title but before the Markdown title.) For example, consider the following Markdown document:

```
Introductory text. Lorem ipsum dolor sit amet, consectetur adipiscing elit.

# Title text

Sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
```

In these cases, the regular expression would match and pull a Markdown title surrounded by text, and cause confusion. So far, *md2conf* manipulated the document as little as possible. I definitely see value in extracting (and removing) the first Markdown heading (unless preceded by other text, as shown above) but then we need a more sophisticated approach not to cause more harm than we do good.

If we manipulate the document, we also need to update unit tests (see directory `tests`) that verify transforming Markdown documents into Confluence Storage Format (CSF) output. CSF reference examples (`*.xml`) should not contain the extracted Markdown heading text.

md2conf/converter.py
giorgek-rfpro: Good spot. You are absolutely right. They cancel each other out. I will update.

md2conf/converter.py
giorgek-rfpro: Yeah, that is a good point. How about only extract the title if it's on the first line? That should be easy to do by checking that first non-empty line starts with `#`, maybe?

hunyadi: In many cases, it could work but there are caveats. For example, some of our documents have several first-level headings in the same Markdown document:

```md
# Basic features
...
# Advanced features
...
# Code examples
...
```

If we took the text from the first heading, we would end up with the Confluence wiki page title *Basic features*, which is definitely not what the entire document is about, it's only one section. However, if there is no other first-level heading, e.g. as in [this document](https://github.com/hunyadi/md2conf/blob/master/README.md), then I believe it's OK. Unfortunately, this would be more complex to implement, and will likely entail scanning the document structure once it's parsed into an XHTML tree.

In a nutshell, I would remain cautious. If there is no obvious great solution, I would rather go with a blank title (as is the case today), which is readily apparent to the user. A wrong guess may be tougher to identify.

hunyadi: I am happy to look into this issue, I don't want to push extra work onto you. In fact, I didn't think of the above use case I mentioned just above when I started reviewing the pull request, and I am grateful that our discussion led to the discovery that we don't have unit tests to cover this special case of several first-level Markdown headings. Whether we choose to pull the very first heading we encounter, or choose to keep it in the document, we still have the issue how we identify the right top-level heading. My current take is that we will need to inspect the document tree, a regular expression won't suffice.

giorgek-rfpro: You are right, the situation is more complicated in general. Adding extra complex logic might be harder to explain to the user. Why having one `#` heading strips it from the document and sets page title and having two `#` headings ends up leaving the headings in place and setting page title as filename instead.

Interestingly, that a similar Java tool uses a separate [CLI argument](https://github.com/md2conf/md2conf?tab=readme-ov-file#title-processing-options) for that.

I would suggest that we need to update [README.md](https://github.com/gkoulin/md2conf/blob/1e03062806e27a08411e73f7355254cafbc05efa/README.md#L136), as it led me to believe this was the current state of the tool. I spend quite a bit of time diving into the code to figure out that this case isn't handled, which led me to this PR 🙂.

hunyadi: As the first step, I implemented a simple table of contents constructed from Markdown headings. If the topmost heading is unique (e.g. only a single first-level heading exists in the document), its text is assigned as the Confluence wiki page title.

I have updated `README.md` to be more explicit about how a Confluence page title is assigned, see the new section *Page title*.

I am still leaving the original Markdown heading in the document, even if unique. We may still choose to implement removing the unique title in a future iteration, as done in the Java tool.

gkoulin: Thanks for this!

nicholasmturner: Hi, I have an issue with v0.3.2 whereby all my pages are created based on the filenames rather than the H1 value. This means my hierarchy isn't honoured because I can't upload multiple index.md files. I'm only currently trying out this tool with some test docs so I just added the titles using the front-matter syntax and this was successful. I'm assuming the above is the reason for it? If so, any ETA on the release containing this change?

hunyadi: The change to extract the unique top-level Markdown heading title as the Confluence document title is not released yet but committed to `master`. Typically, I put several bugfixes and minor features on the same *release train* to reduce noise. Next week seems reasonable for a new minor release. Meanwhile, you can always use the editable install feature in Python and `pip` to run the code as it is in the repository.

PR #117, opened 2025-05-27
Use Confluence metadata to store information about mapping .md files

xxBedy:
* Use Confluence metadata to store information about mapping between the Confluence page and Markdown file instead of modifying markdown files. It's usefull when using md2conf in the automation like GitHub actions.
* Links to folders pointing to index.md or README.md if exists.
* Orphaned pages in Confluence can be removed (--remove-orphans parameter introduced. Default: False), only pages previously created by md2conf are deleted.


hunyadi:
> I understand. Can you merge it to the develop branch - or some other branch, so we can cooperate on it?

Absolutely. Please allow me a few days because I will be out for a conference for the rest of the week.

> How can I run whole test suite to see the results? (I’m working as a Java programer for more then 25 years, so I’m familiar with this ecosystem)

The principles are the same as in Java. The project has unit tests (see folder `tests`): these don't communicate with Confluence but generate Confluence output locally (so-called Confluence Storage Format files, which are essentially an extended version of XHTML) and test the output against reference documents. The project also has integration tests (see folder `integration_tests`): these invoke API requests against a private Confluence space. Unfortunately, the API keys for this Confluence space are not shared at the moment for security reasons (e.g. we don't want bots to inject malicious content).

In general, if something can be covered by unit tests, it should be. However, the functionality you are proposing is specifically about the interface between *md2conf* and Confluence, and will require integration tests. I can definitely contribute but we can't merge this proposal to the mainline branch without covering new functionality. (The existing approach utilizing comment tags in Markdown files is fully covered in tests.)

> I think that we can remove the part that writes changes back to the markup files and add the functionality that will add mappings described in markups files in advance to get mapping from Confluence metadata.

While I definitely like the proposed feature of storing the association information in Confluence, we can't drop the functionality that relies on Markdown comment tags (or remove tags from files). We have (internal) tooling that depends on the presence of these tags, and we don't want to rewrite those tools at the moment. One particular example that comes to mind is that we synchronize from several GitHub repositories to the same Confluence space. This is fully supported with the existing implementation but may be more challenging with the proposed approach (which, as far as I currently understand, might unintentionally delete pages owned by the other project team, or leave orphaned pages). We can probably mitigate these challenges (e.g. use some extra information that represents the root of synchronization) but I don't want to jump into deep water without fully exploring impact.

On the other hand, we can absolutely **extend** how *md2conf* behaves, e.g. add a new command-line switch that allows the association to be stored (1) in the Markdown file (existing behavior), (2) in Confluence page metadata (proposed behavior), or (3) both. This allows seamless adoption.

md2conf/application.py
```diff
@@ -11,7 +11,7 @@
 from pathlib import Path
 from typing import Optional
 
-from .api import ConfluencePage, ConfluenceSession
+from .api import ConfluencePage, ConfluenceSession, LOGGER
```
hunyadi, line 14: This seems an error/oversight. All logged messages in `application.py` should use the `LOGGER` module-level instance (instantiated in the same file).

md2conf/application.py
```diff
@@ -92,21 +93,23 @@ def _get_or_create_page(
...
-            self.api.space_id_to_key(confluence_page.space_id),
-        )
+        page_id = self.get_existing_page_id(self.root_dir, absolute_path)
+
+        if page_id is not None:
+            LOGGER.debug("Retrieving existing page: %s", page_id)
+            confluence_page = self.api.get_page(page_id)
+        else:
+            LOGGER.debug("Creating new page with title: %s", title)
+            confluence_page = self.api.create_page(parent_id.page_id, title, content, self.root_dir, absolute_path)
```
hunyadi, line 111: Would it be possible to merge this functionality with the existing functionality? In other words, a user could associate a document with a page either in the Markdown file or in Confluence metadata. The application could check (and update) both places automatically. This would allow seamless behavior for existing use cases that rely on comment tags in the Markdown file.

md2conf/application.py
```diff
@@ -155,36 +158,24 @@ def _save_document(
...
 
-        content.append(document[index:])
+    def remove_existing_page_from_dict(self, root_dir, absolute_path):
+        path_digest = self.api.get_path_digest(root_dir, absolute_path)
+        if path_digest in self.existing_pages:
+            LOGGER.debug("Removing existing page from dictionary: %s", path_digest)
+            del self.existing_pages[path_digest]
 
-        with open(path, "w", encoding="utf-8") as file:
-            file.write("\n".join(content))
```
hunyadi, line 187: I think we should keep the existing functionality as well for backwards compatibility. If we want to give users more control, we could introduce a command-line switch that allows (or disallows) updating the Markdown document with the assigned Confluence page ID.

md2conf/converter.py
```diff
@@ -454,6 +454,12 @@ def _transform_link(self, anchor: ET._Element) -> Optional[ET._Element]:
                 raise DocumentError(msg)
 
         link_metadata = self.page_metadata.get(absolute_path)
+
+        if link_metadata is None:
+            link_metadata = self.page_metadata.get(absolute_path.joinpath("index.md"))
+        if link_metadata is None:
+            link_metadata = self.page_metadata.get(absolute_path.joinpath("README.md"))
```
hunyadi, line 461: Can you help me understand why we need these extra assignments? Relative links in Markdown files should always point to other Markdown files, links to `index.md` and `README.md` are no exceptions. Relative links should never be pointers to a file system directory.

md2conf/api.py
```diff
@@ -774,23 +803,73 @@ def page_exists(
...
-    def get_or_create_page(self, title: str, parent_id: str) -> ConfluencePage:
-        """
-        Finds a page with the given title, or creates a new page if no such page exists.
+    def get_existing_pages(self, parent_page_id: str) -> dict[str, str]:
+        """Get all existing pages under a parent page"""
 
-        :param title: Page title. Pages in the same Confluence space must have a unique title.
-        :param parent_id: Identifies the parent page for a new child page.
-        """
+        all_data = []
```
hunyadi, line 809: You need to add type information if the type cannot be unambiguously inferred from the context, e.g. `all_data: list[str] = []`. (This is similar to how this works in Java.) In general, `check.sh` or `check.bat` in the project root runs the static type checker, which should validate whether the project code is fully typed.

md2conf/api.py
```diff
@@ -774,23 +803,73 @@ def page_exists(
...
-            return self.create_page(parent_id, title, "")
+        while True:
+            try:
+                query = {
+                    'start': start,
+                    'limit': limit,
+                    'expand': 'metadata.properties.md2conf'
+                }
+
+                url = self._build_url(ConfluenceVersion.VERSION_1, f"/content/{parent_page_id}/child/page")
```
hunyadi, line 823: Looks like we are invoking REST API v1, which is being phased out in favor of REST API v2. Fortunately, v2 API seems to have an [endpoint](https://developer.atlassian.com/cloud/confluence/rest/v2/api-group-content-properties/#api-pages-page-id-properties-get) with similar functionality.

md2conf/api.py
```diff
@@ -711,6 +718,28 @@ def create_page(
...
         )
 
+    def get_path_digest(self, root_dir: Path, absolute_path: Path) -> str:
+        """
+        Calculate a digest for the relative path of a file under the root directory.
+        This is used to uniquely identify the file in Confluence metadata.
+        """
+        relative_path = absolute_path.relative_to(root_dir.parent)
+        filepath_hash = hashlib.md5(relative_path.as_posix().encode("utf-8"))
+        return "".join(f"{c:x}" for c in filepath_hash.digest())
```
hunyadi, line 728: It might be worth thinking through what we want to store as metadata. A digest is a one-way function, and we will still require the Markdown hierarchy at the original source to make sense of it. It's also sensitive to renames and moves, and can easily leads to an orphaned page. On the other hand, many of our Markdown files are stored in Git repositories. A direct URL to the Markdown document in GitHub/GitLab might be an option to consider. Even if the URL changes, Git history allows us to track down where Git and Confluence got misaligned.

md2conf/api.py
```diff
@@ -774,23 +803,73 @@ def page_exists(
...
+                print(f"Error during API request: {e}")
+                return []
+
+            except json.JSONDecodeError as e:
+                print(f"Error decoding JSON response: {e}")
+                return []
+
+            except Exception as e:
+                print(f"An unexpected error occurred: {e}")
+                return []
```
hunyadi, line 855: Errors should be logged with `LOGGER`, which has a dedicated member function to log exceptions.

hunyadi:
I am closing this pull request due to significant skew from the state on the branch `master`. However, I believe the core idea has been incorporated into the latest state by completely separating the indexing and local/remote synchronization phases. In the future, local/remote synchronization can look not only at page ID and title inferred from Markdown but also other characteristics (e.g. one-to-one correspondence between child documents and pages). This can make synchronization independent of IDs maintained in Markdown comments but without the need to keep cryptic digests in Confluence page metadata.

Issue #128, opened 2025-07-02
Archived pages are causing issues


fcusson:
When trying to add a page that uses the same name as an archived page, this is failing stating that `archived` is not a valid ConfluenceStatus.

full traceback here:

```python
Traceback (most recent call last):
  File "/home/fcusson/Sources/work/tableau/.venv/bin/md2conf", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/fcusson/Sources/work/tableau/.venv/lib64/python3.11/site-packages/md2conf/__main__.py", line 252, in main
    ).process(args.mdpath)
      ^^^^^^^^^^^^^^^^^^^^
  File "/home/fcusson/Sources/work/tableau/.venv/lib64/python3.11/site-packages/md2conf/processor.py", line 280, in process
    self.process_directory(path)
  File "/home/fcusson/Sources/work/tableau/.venv/lib64/python3.11/site-packages/md2conf/processor.py", line 299, in process_directory
    self.factory.create(root_dir).process_directory(local_dir)
  File "/home/fcusson/Sources/work/tableau/.venv/lib64/python3.11/site-packages/md2conf/processor.py", line 110, in process_directory
    self._synchronize_page(path, ConfluencePageID(metadata.page_id))
  File "/home/fcusson/Sources/work/tableau/.venv/lib64/python3.11/site-packages/md2conf/processor.py", line 137, in _synchronize_page
    self._update_page(page_id, document, path)
  File "/home/fcusson/Sources/work/tableau/.venv/lib64/python3.11/site-packages/md2conf/application.py", line 153, in _update_page
    conflicting_page_id = self.api.page_exists(
                          ^^^^^^^^^^^^^^^^^^^^^
  File "/home/fcusson/Sources/work/tableau/.venv/lib64/python3.11/site-packages/md2conf/api.py", line 898, in page_exists
    results = _json_to_object(list[ConfluencePageProperties], data["results"])
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fcusson/Sources/work/tableau/.venv/lib64/python3.11/site-packages/md2conf/api.py", line 47, in _json_to_object
    return json_to_object(typ, data, options=DeserializerOptions(skip_unassigned=True))
[...]

PR #129, opened 2025-07-02
added a filter for archived pages

fcusson:
Potential fix for #128. Enables filtering of archived pages when checking if a page exist based on title.

### DIFF
--- a/md2conf/api.py
+++ b/md2conf/api.py
@@ -901,9 +901,6 @@ class ConfluenceSession:
         data = typing.cast(dict[str, JsonType], response.json())
         results = _json_to_object(list[ConfluencePageProperties], data["results"])
 
-        # filter out archived pages
-        results = [x for x in results if x.status != ConfluenceStatus.ARCHIVED]
-
         if len(results) == 1:
             return results[0].id
         else:
--- a/md2conf/api.py
+++ b/md2conf/api.py
@@ -24,7 +24,7 @@ from strong_typing.serialization import DeserializerOptions, json_dump_string, j
 
 from .converter import ParseError, sanitize_confluence
 from .metadata import ConfluenceSiteMetadata
-from .properties import ArgumentError, ConfluenceConnectionProperties, ConfluenceError, PageError
+from .properties import ArgumentError, ConfluenceConnectionProperties, ConfluenceError, PageError,
 
 T = TypeVar("T")
 
@@ -48,7 +48,8 @@ def build_url(base_url: str, query: Optional[dict[str, str]] = None) -> str:
     if fragment:
         raise ValueError("expected: url with no fragment")
 
-    url_parts = (scheme, netloc, path, None, urlencode(query) if query else None, None)
+    url_parts = (scheme, netloc, path, None,
+                 urlencode(query) if query else None, None)
     return urlunparse(url_parts)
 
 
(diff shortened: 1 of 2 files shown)


md2conf/api.py
```diff
@@ -95,6 +96,7 @@ class ConfluenceRepresentation(enum.Enum):
 class ConfluenceStatus(enum.Enum):
     CURRENT = "current"
     DRAFT = "draft"
+    ARCHIVED = "archived"
```
hunyadi, line 98: Thanks for adding this missing item to the list of statuses that *md2conf* recognizes! Can you please roll back the other code formatting changes? I want to make sure the change is as small as it needs to be, and formatting changes don't obscure the real change. Thanks in advance!

fcusson: Hi @hunyadi, sorry about that, I must have left my auto formatting on when making the change. I'll revert that

fcusson: formatting changes have been removed.

md2conf/api.py
```diff
@@ -900,6 +901,9 @@ def page_exists(
         data = typing.cast(dict[str, JsonType], response.json())
         results = _json_to_object(list[ConfluencePageProperties], data["results"])
 
+        # filter out archived pages
+        results = [x for x in results if x.status != ConfluenceStatus.ARCHIVED]
```
hunyadi: One concern I have here is that the function `page_exists` is used to verify whether the page title is unique. Does Confluence allow you to assign a title to a page which conflicts with the title of an archived page? If not, we will need to alter how `page_exists` is invoked (e.g. raise an exception when synchronization would cause a conflict with an archived page). (I don't think "un-archiving" a page, and then updating it would be the right approach. This feels intrusive. On the other hand, I don't think Confluence allows updating archived pages.)

I am happy to take these extra tasks on myself if you don't want to keep working on this issue. In this case, please keep only the added new value for the set of values that enumeration type has.

fcusson: good point, I tested and confluence does seem to block the creation of a page with same name as an archive page. I though I was able to do it in the past, but must have misremembered.

I agree, un-archiving would be too intrusive and could run the risk of wiping an important archive by accident if the user didn't know this would happen. 

Do you have a preexeisting error already that I could use or would I need to create an exception for that? If we need to create, any parent that could be used?

hunyadi: `page_exists` is invoked in two contexts.

In `application.py`, it's used to check for the presence of a conflicting title. If you don't modify the behavior of `page_exists`, it is already doing the right thing, emitting a message to the log.

In the function `get_or_create_page`, there is a conditional that branches on whether it finds a page with a matching title. If a match is found, the page is returned, otherwise a new page is created. Again, if you don't modify the body of `page_exists`, the implementation is OK. The problem is going to be triggered in the caller function `_synchronize_subtree` in `application.py`. Here, you will need to add an exception for an archived page, which cannot be updated:

```
    def _synchronize_subtree(self, node: DocumentNode, parent_id: ConfluencePageID) -> None:
            ...
            # look up page by title
            page = self.api.get_or_create_page(node.title, parent_id.page_id)
            if page.status is ConfluenceStatus.ARCHIVED:
                raise PageError(f"unable to update archived page with ID {page.id}")
            update = True
```

fcusson: made the changes as discussed. Please tell me if you have any other comments or improvements for the PR.

PR #154, opened 2025-08-25
Add support for configuring Mermaid diagram scale

### LINKED ISSUE #153: Feature Request: Add support for configuring Mermaid diagram scale
## Problem Description
Hi,

I've been using md2conf to automate documentation publication, which provides excellent flexibility for generating Mermaid diagrams independently of GitLab's embedded version. However, I'm struggling with controlling the size of rendered images, which are strongly tied to the diagram's inherent size, leading to disproportionate scaling in some cases.

It's quite challenging to generate appropriately sized Mermaid diagrams that are clearly displayed in Confluence. Some diagrams render too small to be readable, while others become excessively large. Issue #135 already highlighted the need for means to customize image sizes.

## Current Workaround Limitations
Currently, the only way to influence diagram size is through mermaid-cli's default scale parameter (hardcoded to "2" in mermaid.py), which applies uniformly to all diagrams regardless of their complexity or content requirements.

## Proposed Solution
I was thinking to provide `width` and `height` as input parameters and I ended up fining that making the existing `scale` parameter flexible is probably good enough.
A possible solution would be to add support for extracting the scale parameter from Mermaid's YAML [front matter configuration](https://mermaid.js.org/config/configuration.html) and passing it to mermaid-cli's --scale option.

> Note: [Directives](https://mermaid.js.org/config/directives) are deprecated

**Example usage:**
```
---
title: Tiny flow diagram
config:
  scale: 1
---
flowchart LR
    A[Component A] --> B[Component B]
    B --> C[Component C]
```

Thoughts?

rpioger:
Leverage mermaid front-matter to drive the '--scale' property of the mermaid CLI.
closes #153

### DIFF
--- a/md2conf/converter.py
+++ b/md2conf/converter.py
@@ -21,7 +21,7 @@ from urllib.parse import ParseResult, quote_plus, urlparse
 
 import lxml.etree as ET
 from strong_typing.core import JsonType
-from strong_typing.deserializer import JsonTypeError
+from strong_typing.exception import JsonTypeError
 
 from . import drawio, mermaid
 from .collection import ConfluencePageCollection
@@ -32,8 +32,9 @@ from .environment import PageError
 from .extra import override, path_relative_to
 from .latex import get_png_dimensions, remove_png_chunks, render_latex
 from .markdown import markdown_to_html
+from .mermaid import MermaidConfigProperties
 from .metadata import ConfluenceSiteMetadata
-from .scanner import ScannedDocument, Scanner
+from .scanner import MermaidScanner, ScannedDocument, Scanner
 from .toc import TableOfContentsBuilder
 from .uri import is_absolute_url, to_uuid_urn
 from .xml import element_to_text
@@ -390,7 +391,6 @@ class ConfluenceStorageFormatConverter(NodeVisitor):
         self.embedded_files = {}
         self.site_metadata = site_metadata
         self.page_metadata = page_metadata
-        self.scanner = Scanner()  # Create once, reuse multiple times
 
     def _transform_heading(self, heading: ET._Element) -> None:
         """
@@ -829,14 +829,14 @@ class ConfluenceStorageFormatConverter(NodeVisitor):
             AC_ELEM("plain-text-body", ET.CDATA(content)),
         )
 
-    def _extract_mermaid_scale(self, content: str) -> Optional[float]:
+    def _extract_mermaid_config(self, content: str) -> Optional[MermaidConfigProperties]:
         """Extract scale from Mermaid YAML front matter configuration."""
         try:
-            properties = self.scanner.fetch_mermaid_properties(content)
+            properties = MermaidScanner().read(content)
+            return properties.config
         except JsonTypeError as ex:
             LOGGER.warning("Failed to extract Mermaid properties: %s", ex)
             return None
-        return properties.config.scale
 
     def _transform_external_mermaid(self, absolute_path: Path, attrs: ImageAttributes) -> ET._Element:
         "Emits Confluence Storage Format XHTML for a Mermaid diagram read from an external file."
@@ -848,8 +848,8 @@ class ConfluenceStorageFormatConverter(NodeVisitor):
         if self.options.render_mermaid:
             with open(absolute_path, "r", encoding="utf-8") as f:
                 content = f.read()
-            scale = self._extract_mermaid_scale(content)
-            image_data = mermaid.render_diagram(content, self.options.diagram_output_format, scale=scale)
+            config = self._extract_mermaid_config(content)
+            image_data = mermaid.render_diagram(content, self.options.diagram_output_format, config=config)
             image_filename = attachment_name(relative_path.with_suffix(f".{self.options.diagram_output_format}"))
             self.embedded_files[image_filename] = EmbeddedFileData(image_data, attrs.alt)
             return self._create_attached_image(image_filename, attrs)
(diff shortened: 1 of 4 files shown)


md2conf/converter.py
```diff
@@ -827,6 +828,37 @@ def _transform_code_block(self, code: ET._Element) -> ET._Element:
             AC_ELEM("plain-text-body", ET.CDATA(content)),
         )
 
+    def _extract_mermaid_dimensions(self, content: str) -> Optional[str]:
+        """Extract scale from Mermaid YAML front matter configuration."""
+        
+        # Look for YAML front matter pattern (--- ... ---)
+        yaml_pattern = r'^---\s*\n(.*?)\n---\s*\n'
+        match = re.search(yaml_pattern, content, re.DOTALL | re.MULTILINE)
```
hunyadi: Can you double-check this regular expression? `\s*` also matches `\n`, and `^` anchors at the beginning of the line for `re.MULTILINE` (e.g. you could use it for matching the closing `---`).

The project already has a function to extract front-matter from the beginning of a Markdown file. We could consider calling the same function from both places.

md2conf/converter.py
```diff
@@ -827,6 +828,37 @@ def _transform_code_block(self, code: ET._Element) -> ET._Element:
...
+            
+            # Check for scale in config section
+            if isinstance(config, dict) and 'config' in config:
+                config_section = config['config']
+                if isinstance(config_section, dict):
+                    scale = config_section.get('scale')
+            
+            # Convert to string if it's a number
+            if isinstance(scale, (int, float)):
+                scale = str(scale)
```
hunyadi: While this is OK, the project uses data-classes for capturing properties/settings, and type-safe conversion from JSON to a `dataclass` instance, see the Python source file `scanner.py`. I understand this might be an overkill for a single property `scale` but the structured approach scales well as we add more properties of interest, and also promotes applying the same techniques across the project (e.g. we parse front-matter in the same way, no matter if it is Markdown front-matter or Mermaid front-matter).

md2conf/converter.py
```diff
@@ -827,6 +828,37 @@ def _transform_code_block(self, code: ET._Element) -> ET._Element:
...
+                
+            return scale
+            
+        except (yaml.YAMLError, KeyError, ValueError):
+            return None
```
hunyadi: I understand the `except` clause for `YAMLError` but can you give some hint as to why we need to catch `KeyError` and `ValueError`? In general, I am afraid to catch generic exception classes and silently ignoring them because this may obscure real programming errors (e.g. indexing a non-existing element in a dictionary).

Even if we catch the error, we should log an informational error message. This calls the user's attention that they are dealing with a malformed Mermaid file.

md2conf/mermaid.py
```diff
@@ -44,7 +44,7 @@ def has_mmdc() -> bool:
     return shutil.which(executable) is not None
 
 
-def render_diagram(source: str, output_format: Literal["png", "svg"] = "png") -> bytes:
+def render_diagram(source: str, output_format: Literal["png", "svg"] = "png", scale: str = None) -> bytes:
```
hunyadi: I think scale should be a `Decimal` to keep original precision yet indicate that this is a number rather than an arbitrary string such as `"larger"`.

md2conf/scanner.py
```diff
@@ -155,3 +166,31 @@ def read(self, absolute_path: Path) -> ScannedDocument:
...
+            # Check for scale in `config` section
+            if isinstance(properties, dict) and 'config' in properties:
+                config_section = properties['config']
+                if isinstance(config_section, dict):
+                    scale = config_section.get('scale')
+                    if scale is not None and not isinstance(scale, (int, float)):
+                        raise ValueError("'scale' property must be a decimal number.")
+        return MermaidRenderProperties(
+            scale=scale
+        )
```
hunyadi: Thanks for making these changes, they have already made the pull request much more compact!

A further opportunity I see is to parse extracted front-matter data the same way as already done for Markdown front-matter. The key point is the following call:

```py
p = _json_to_object(DocumentProperties, data)
```

Notice how this is both making sure that the front-matter adheres to the expected structure, and populates an object of type `DocumentProperties` with the right data. You can use the same function call to populate a `MermaidRenderProperties` which has a field `config`, which would be of type `MermaidOutputConfiguration`, which would in turn have a field called `scale` basically mimicking the front-matter JSON/YAML structure. This would accomplish the same as the highlighted code, but hide all implementation detail of type checks and conversions in the call to `_json_to_object`.

Originally, I used the same approach you do in the highlighted code. Unfortunately, the code quickly became bloated as I added more properties to extract. However, putting everything under `DocumentProperties` kept things clean.

hunyadi (commented):
Thanks for the updates, I appreciate all the work you put into making this more compact. Feel free to let me know if you need help or I should take over. After a few minor corrections, I am more than happy to merge.

md2conf/scanner.py
hunyadi: One huge advantage of having a `MermaidRenderProperties` (or, with a less specific name, `MermaidProperties`) would be that I could add the `title` field as shown on the [Configuration](https://mermaid.js.org/config/configuration.html) page. Earlier, I struggled to give a Mermaid figure a meaningful title when the [Mermaid diagram was defined inline with fenced code blocks](https://github.com/hunyadi/md2conf/blob/master/README.md#mermaid-diagrams). However, if Mermaid supports front-matter, and we extract `title` from the front-matter, we could finally have real figure captions! This gives more incentive to have a `dataclass` like `DocumentProperties`.

PR #173, opened 2025-12-01
Allow specifying links in absolute paths

gpollo:
Thank you for making this tool. We're experimenting with this tool to export our Markdown documentation to Confluence.

### Problem

We have the following files:
```
index.md
tools.md
images/
    apple.svg
    banana.svg
subdir/
    index.md
    maple_sirup.md
```

Inside `maple_sirup.md`, we have links that are absolute (relative to root of the file hierarchy).

```markdown
# My File

[tool list](/tools.md)

![banana image](/images/banana.svg)
```

Currently, `md2conf` will fail to convert this file, because it will actually try to get the image from `/images/banana.svg`.

### Solution

Root links should be fetched from the parent directory of the root page.

If the root page is `/home/danbigbras/workspace/index.md`, then

* `/tools.md` → `/home/danbigbras/workspace/tools.md`.
* `/images/banana.svg` → `/home/danbigbras/workspace/images/banana.svg`.

### Rational

That's how VS Code works when rendering Markdown. I assume Github Markdown also works that way. Well, they're relative to the root folder opened or the repository.

### DIFF
--- a/md2conf/converter.py
+++ b/md2conf/converter.py
@@ -81,6 +81,12 @@ def is_directory_within(absolute_path: Path, base_path: Path) -> bool:
     return absolute_path.as_posix().startswith(base_path.as_posix())
 
 
+def fix_absolute_path(path: Path, root_path: Path) -> Path:
+    "Make absolute path relative to another root path."
+
+    return root_path / path.relative_to(path.root)
+
+
 def encode_title(text: str) -> str:
     "Converts a title string such that it is safe to embed into a Confluence URL."
 
@@ -545,9 +551,11 @@ class ConfluenceStorageFormatConverter(NodeVisitor):
 
         # discard original value: relative links always require transformation
         anchor.attrib.pop("href")
-
-        # convert the relative URL to absolute path based on the base path value
-        absolute_path = (self.base_dir / relative_url.path).resolve()
+        if relative_url.path.startswith("/"):
+            absolute_path = fix_absolute_path(path=Path(relative_url.path), root_path=self.root_dir).resolve()
+        else:
+            # convert the relative URL to absolute path based on the base path value
+            absolute_path = (self.base_dir / relative_url.path).resolve()
 
         # look up the absolute path in the page metadata dictionary to discover the relative path within Confluence that should be used
         if not is_directory_within(absolute_path, self.root_dir):
@@ -716,8 +724,11 @@ class ConfluenceStorageFormatConverter(NodeVisitor):
     def _verify_image_path(self, path: Path) -> Path | None:
         "Checks whether an image path is safe to use."
 
-        # resolve relative path into absolute path w.r.t. base dir
-        absolute_path = (self.base_dir / path).resolve()
+        if path.is_absolute():
+            absolute_path = fix_absolute_path(path=path, root_path=self.root_dir).resolve()
+        else:
+            # resolve relative path into absolute path w.r.t. base dir
+            absolute_path = (self.base_dir / path).resolve()
 
         if not absolute_path.exists():
             self._warn_or_raise(f"path to image {path} does not exist")


md2conf/converter.py
```diff
@@ -716,8 +721,13 @@ def _warn_or_raise(self, msg: str) -> None:
...
 
-        # resolve relative path into absolute path w.r.t. base dir
-        absolute_path = (self.base_dir / path).resolve()
+        if os.path.isabs(path):
+            # convert the absolute URL to absolute path relative on root directory
+            without_first_part = "/".join(Path(path).parts[1:])
+            absolute_path = (self.root_dir / without_first_part).resolve()
+        else:
+            # resolve relative path into absolute path w.r.t. base dir
+            absolute_path = (self.base_dir / path).resolve()
```
hunyadi: This looks to be a repetition of the changes above. I would recommend moving the code into a function body, and then calling the function from both places. Keeping internal logic at a single place helps future maintainers.

md2conf/converter.py
```diff
@@ -546,8 +546,13 @@ def _transform_link(self, anchor: ElementType) -> ElementType | None:
         # discard original value: relative links always require transformation
         anchor.attrib.pop("href")
 
-        # convert the relative URL to absolute path based on the base path value
-        absolute_path = (self.base_dir / relative_url.path).resolve()
+        if os.path.isabs(relative_url.path):
+            # convert the absolute URL to absolute path relative on root directory
+            without_first_part = "/".join(Path(relative_url.path).parts[1:])
```
hunyadi: I am not sure I understand the sequence of operations in this line. Wouldn't it suffice to drop the leading `/`, and then concatenate the path with `self.root_dir` rather than `self.base_dir`?

md2conf/converter.py
```diff
@@ -546,8 +546,13 @@ def _transform_link(self, anchor: ElementType) -> ElementType | None:
         # discard original value: relative links always require transformation
         anchor.attrib.pop("href")
 
-        # convert the relative URL to absolute path based on the base path value
-        absolute_path = (self.base_dir / relative_url.path).resolve()
+        if os.path.isabs(relative_url.path):
```
hunyadi: For consistency, we might want to consider [Path.is_absolute](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.is_absolute), since `pathlib` is typically how the *md2conf* library manipulates paths. Exceptions exist, in cases where some `pathlib` methods are not supported in older Python versions.

Even simpler, we can check if the string starts with a `/`. This path comes from a URL (return value of `urlparse`), and always has `/` as the directory separator, we don't need to differentiate between Linux and Windows.

hunyadi (changes requested):
Thank you for the contribution!

I have made some comments to help future code maintainers. Let me know if you lack the time or capacity to make these changes, and I will jump in as necessary.

md2conf/converter.py
gpollo: I agree, I thought they were in different files and I was lazy.

I moved common logic into a `fix_absolute_path` function.

md2conf/converter.py
gpollo: Yes, dropping the leading `/` would work but I was worried it wouldn't work for Windows paths. That said, from your other comments, we shouldn't need to worry for Windows paths.

I still kept it for now as it's make more sense to me.

md2conf/converter.py
gpollo: Removed duplication.

md2conf/converter.py
```diff
@@ -81,6 +81,16 @@ def is_directory_within(absolute_path: Path, base_path: Path) -> bool:
     return absolute_path.as_posix().startswith(base_path.as_posix())
 
 
+def fix_absolute_path(path: Path | str, root_path: Path) -> Path:
```
hunyadi: Since one of the callers already has a `Path` object when invoking `fix_absolute_path`, it may make sense to move the conversion from `str` to `Path` to outside of `fix_absolute_path`, and make it take only `Path` rather than `Path | str`. This might make the signature more expressive as to the intended use.

md2conf/converter.py
```diff
@@ -81,6 +81,16 @@ def is_directory_within(absolute_path: Path, base_path: Path) -> bool:
...
+    "Make absolute path relative to another root path."
+
+    if isinstance(path, str):
+        path = Path(path)
+
+    without_first_part = "/".join(path.parts[1:])
+    return root_path / without_first_part
```
hunyadi: Unfortunately, this still feels magic to me. I understand that `path.parts[0]` is going to evaluate to `/`, and the expression will eventually yield what we need but the whole expression remains hard to grasp. Instead, I would write something like

```
return root_path / path.relative_to(path.root)
```

Specifically, the following succeeds (both on Linux and Windows):

```
p = Path("/path/to/image.png")
assert p.relative_to(p.root).as_posix() == "path/to/image.png"
```

If we use the combination of `Path.relative_to` and `Path.root`, we might not need the function `fix_absolute_path` at all because the expression becomes so simple.

md2conf/converter.py
gpollo: You're right, I agree. Updated.

gpollo:
I agree with your suggestions, they're cleaner. Updated. Thanks!

PR #177, opened 2025-12-03
fix(api): add retry logic for get_page to handle eventual consistency

fmosca:
Confluence Cloud API exhibits eventual consistency behavior where a newly created page may not be immediately available for retrieval via GET request. This causes 404 errors when the tool attempts to fetch the page immediately after creation.

This change adds retry logic to the get_page() method:
- Retries up to 3 times on 404 errors (configurable via 'retries' parameter)
- Uses exponential backoff: 1s, 2s, 4s delays between attempts
- Only retries on 404 status code, other errors are raised immediately
- Logs retry attempts at DEBUG level for troubleshooting

This resolves race conditions when publishing new pages to Confluence Cloud where the page creation succeeds but the subsequent content update fails with a 404 because the page isn't yet queryable.

### DIFF
--- a/md2conf/api.py
+++ b/md2conf/api.py
@@ -11,7 +11,9 @@ import enum
 import io
 import logging
 import mimetypes
+import random
 import ssl
+import time
 import typing
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


tests/test_api_retry.py
```diff
@@ -0,0 +1,43 @@
...
+        # Check the method signature includes retry parameters
+        sig = inspect.signature(ConfluenceSession.get_page)
+        params = list(sig.parameters.keys())
+
+        self.assertIn("retries", params)
+        self.assertIn("retry_delay", params)
+
+        # Verify default values
+        self.assertEqual(sig.parameters["retries"].default, 3)
+        self.assertEqual(sig.parameters["retry_delay"].default, 1.0)
```
hunyadi: I don't think we need this test, do we? All new parameters have default values, the signature is backward compatible.

tests/test_api_retry.py
```diff
@@ -0,0 +1,43 @@
...
+        """Verify the exponential backoff delay calculation."""
+        # delay = retry_delay * (2 ** attempt)
+        retry_delay = 1.0
+
+        # Attempt 0: 1.0 * 2^0 = 1.0
+        self.assertEqual(retry_delay * (2**0), 1.0)
+        # Attempt 1: 1.0 * 2^1 = 2.0
+        self.assertEqual(retry_delay * (2**1), 2.0)
+        # Attempt 2: 1.0 * 2^2 = 4.0
+        self.assertEqual(retry_delay * (2**2), 4.0)
```
hunyadi: I am afraid I am not sure what value this test case is adding. It's not calling any functions, it's essentially validating the exponentiation operator, which is part of core Python. I would suggest removing it.

md2conf/api.py
```diff
@@ -839,16 +840,38 @@ def get_page_properties_by_title(
...
         path = f"/pages/{page_id}"
-        return self._get(ConfluenceVersion.VERSION_2, path, ConfluencePage, query={"body-format": "storage"})
+        last_error: requests.HTTPError | None = None
+
+        for attempt in range(retries + 1):
+            try:
+                return self._get(ConfluenceVersion.VERSION_2, path, ConfluencePage, query={"body-format": "storage"})
+            except requests.HTTPError as e:
+                if e.response is not None and e.response.status_code == 404 and attempt < retries:
+                    delay = retry_delay * (2**attempt)
```
hunyadi: As far as I am aware, the industry practice is to add a small random delta (e.g. within 1 sec) such that retry requests (by independent clients) don't group at exact moments in time.

tests/test_api_retry.py
fmosca: yes, sorry, this test file got pushed before I had a chance to do a pre-review myself. 🤦‍♂️

md2conf/api.py
fmosca: good point, I was thinking of (my) single user, single page case where we don't do multiple get_page concurrently, but that's certainly possible in more complex use cases.

Issue #179, opened 2025-12-12
[Feature] Better mermaid->svg support


brettschneider:
### Describe your idea

In trying to use this with mermaid diagrams rendering to svg, if they are larger diagrams, the resulting svg files and/or `ac:` tags didn't always result in a good page view experience:
- sometimes no image at all
- sometimes so small it's invisible (1px x 1px)
- sometimes so massive it overwhelms the rest of the page content

Would be nice if there was some better handling and even a cli arg to specify a max width of the image displayed in the page.

PR #180, opened 2025-12-12
Better mermaid -> svg support (issue #179)

brettschneider:
This PR addresses the issue I just created.  My client loves the tool, just we had some minor issues with svg and their larger diagrams.  Thanks for all your work!

### DIFF
--- a/md2conf/svg.py
+++ b/md2conf/svg.py
@@ -258,27 +258,28 @@ def _parse_svg_length(value: str) -> int | None:
         return None
 
     # Convert to pixels (assuming 96 DPI, 16px base font)
-    if unit is None or unit.lower() == "px":
-        pixels = num
-    elif unit.lower() == "pt":
-        pixels = num * 96 / 72  # 1pt = 1/72 inch
-    elif unit.lower() == "in":
-        pixels = num * 96
-    elif unit.lower() == "cm":
-        pixels = num * 96 / 2.54
-    elif unit.lower() == "mm":
-        pixels = num * 96 / 25.4
-    elif unit.lower() == "pc":
-        pixels = num * 96 / 6  # 1pc = 12pt = 1/6 inch
-    elif unit.lower() == "em":
-        pixels = num * 16  # assume 16px base font
-    elif unit.lower() == "ex":
-        pixels = num * 8  # assume ex ≈ 0.5em
-    elif unit == "%":
-        # Percentage values can't be resolved without a container; skip
-        return None
-    else:
-        return None
+    match unit.lower() if unit else None:
+        case None | "px":
+            pixels = num
+        case "pt":
+            pixels = num * 96 / 72  # 1pt = 1/72 inch
+        case "in":
+            pixels = num * 96
+        case "cm":
+            pixels = num * 96 / 2.54
+        case "mm":
+            pixels = num * 96 / 25.4
+        case "pc":
+            pixels = num * 96 / 6  # 1pc = 12pt = 1/6 inch
+        case "em":
+            pixels = num * 16  # assume 16px base font
+        case "ex":
+            pixels = num * 8  # assume ex ≈ 0.5em
+        case "%":
+            # Percentage values can't be resolved without a container; skip
+            return None
+        case _:
+            return None
 
     return int(round(pixels))
 


README.md
```diff
@@ -507,46 +507,78 @@ $ python3 -m md2conf --help
...
-                        Alignment for block-level images and formulas (default: 'center').
-  --use-panel           Transform admonitions and alerts into a Confluence custom panel.
+                        Alignment for block-level images and formulas
+                        (default: 'center').
+  --use-panel           Transform admonitions and alerts into a Confluence
+                        custom panel.
+  --max-image-width MAX_IMAGE_WIDTH
+                        Maximum display width for images in pixels. Images
+                        wider than this will be scaled down for display while
+                        preserving the original size for full-size viewing.
```
hunyadi: Many of these lines appear to be reformatting. Can you remove the phantom changes that pertain to different placement of a line break?

md2conf/converter.py
```diff
@@ -309,7 +312,9 @@ def as_dict(self) -> dict[str, str]:
                 attributes[AC_ATTR("original-height")] = str(self.height)
             if self.width is not None:
                 attributes[AC_ATTR("custom-width")] = "true"
-                attributes[AC_ATTR("width")] = str(self.width)
+                # Use display_width if set, otherwise use natural width
+                effective_width = self.display_width if self.display_width is not None else self.width
```
hunyadi: Unless `display_width` can legitimately take a value of 0, it may be simpler to write:

```
effective_width = self.display_width or self.width
```

md2conf/converter.py
```diff
@@ -749,6 +776,21 @@ def _transform_attached_image(self, absolute_path: Path, attrs: ImageAttributes)
             if png_file.exists():
                 absolute_path = png_file
 
+        # infer SVG dimensions if not already specified
+        if absolute_path.suffix == ".svg" and attrs.width is None and attrs.height is None:
+            svg_width, svg_height = get_svg_dimensions(absolute_path)
+            if svg_width is not None or svg_height is not None:
```
hunyadi: Can you double-check if you specifically need to check for `svg_width` ***or*** `svg_height`? I would assume you need both, or at least width. (HTML rendering is width-based.)

md2conf/svg.py
```diff
@@ -0,0 +1,256 @@
...
+        text = data.decode("utf-8")
+
+        # Skip SVGs with foreignObject - Confluence has issues rendering
+        # foreignObject content when explicit width/height are set on the SVG
+        if "<foreignObject" in text:
+            LOGGER.debug("Skipping dimension fix for SVG with foreignObject elements")
+            return data
+
+        # Extract the SVG opening tag
+        svg_tag_match = re.search(r"<svg[^>]+>", text)
```
hunyadi: While this is likely not causing any issues, you technically need a word boundary marker after `svg` in the regular expression to avoid matching tags that start with (but not equal to) `svg`.

md2conf/svg.py
```diff
@@ -0,0 +1,256 @@
...
+        svg_tag = svg_tag_match.group(0)
+
+        # Check if we need to fix (has width="100%" or similar percentage)
+        if 'width="100%"' not in svg_tag:
+            # Check if it already has a valid numeric width
+            if re.search(r'\swidth="\d+(?:\.\d+)?"', svg_tag):
+                return data  # Already has numeric width
+
+        # Extract viewBox dimensions
+        viewbox_match = re.search(r'viewBox="([^"]+)"', svg_tag)
```
hunyadi: Like in the case of matching for the attribute `width`, you probably need a word boundary marker before the literal `viewBox`. Since XML mandates putting at least one whitespace before attributes, a `\s` may also be OK.

md2conf/svg.py
```diff
@@ -0,0 +1,256 @@
...
+            return data
+
+        vb_width, vb_height = _parse_viewbox(viewbox_match.group(1))
+        if vb_width is None or vb_height is None:
+            return data
+
+        # Replace width="100%" with explicit width
+        text = re.sub(r'(<svg[^>]*)\swidth="100%"', rf'\1 width="{vb_width}"', text)
+
+        # Re-extract the SVG tag after width replacement
```
hunyadi: This approach seems somewhat complicated. I understand the motivation to use regular expressions instead of `lxml` to avoid accidentally updating parts in the rest of the document (e.g. `CDATA`, which `lxml` doesn't exactly reproduce in a deserialization and re-serialization cycle). However, we could simply extract the root node with all of its attributes but none of its content, make changes there (with XML tooling or otherwise) and then plug in the start tag of the updated root node `<svg width="..." ...>` in the original document. This can help avoid some of the string manipulation below (e.g. rescanning the same SVG document multiple times), and make this part easier to read.

md2conf/svg.py
```diff
@@ -0,0 +1,256 @@
...
+        pixels = num * 96 / 6  # 1pc = 12pt = 1/6 inch
+    elif unit.lower() == "em":
+        pixels = num * 16  # assume 16px base font
+    elif unit.lower() == "ex":
+        pixels = num * 8  # assume ex ≈ 0.5em
+    elif unit == "%":
+        # Percentage values can't be resolved without a container; skip
+        return None
+    else:
+        return None
```
hunyadi: *md2conf* versions 0.5.0 and later require Python 3.10 or later, which has the `match` statement. This long series of `elif` might be easier to write (and read) as a `match` on `unit.lower()`.

README.md
brettschneider: Done

md2conf/converter.py
brettschneider: Done

md2conf/converter.py
brettschneider: Done

md2conf/svg.py
brettschneider: Done

md2conf/svg.py
brettschneider: Done

md2conf/svg.py
brettschneider: Done

md2conf/svg.py
brettschneider: Done

PR #181, opened 2025-12-15
feat(converter): add limited template option for generated by text

baptman21:
## Feature request

Currently, it's possible to have a generated header at the start of the document.

The use case I would like to have is the ability to have a link to edit the files in the repository quickly "Click here to edit".
This is a similar behavior in a lot of documentation tools, to refer back to the Github repository.

## Implementation

Simply give the ability to template the `generated-by` text with `filename` and `filepath`. This would allow users to set a generated by message like: `Edit the file: https://github.com/user/repo/edit/master/{filepath}`.

## Tests

Added a test for the templating and the templating of an unexpected variable.

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,12 +172,14 @@ Alternatively, use the `--generated-by GENERATED_BY` option. The tag takes prece
 The generated-by text can also be templated with the following variables:
 
 - `%{filename}`: the name of the Markdown file
-- `%{filepath}`: the path of the Markdown file relative to the source directory. When [publishing a single page](#publishing-a-single-page), it's the same as the filename.
+- `%{filepath}`: the path of the Markdown file relative to the *source root*
+
+When publishing a directory hierarchy, the *source root* is the directory in which *md2conf* is launched. When publishing a single file, this is the directory in which the Markdown file resides.
 
 It can be used with the CLI `--generated-by` option or directly in the files:
 
 ```markdown
-<!-- generated-by: Do not edit! Check out this file %{filepath} in the repo-->
+<!-- generated-by: Do not edit! Check out the file %{filepath} in the repo -->
 ```
 
 ### Publishing a single page
@@ -561,7 +563,7 @@ options:
                         Alignment for block-level images and formulas (default: 'center').
   --use-panel           Transform admonitions and alerts into a Confluence custom panel.
   --max-image-width MAX_IMAGE_WIDTH
-                        Maximum display width for images in pixels. Images wider than this will be scaled down for display while preserving the original size for full-size viewing.
+                        Maximum display width for images [px]. Wider images are scaled down for page display. Original size kept for full-size viewing.
 ```
 
 ### Confluence REST API v1 vs. v2


md2conf/converter.py
```diff
@@ -43,6 +43,24 @@
...
+
+    :param template: The template string with placeholders
+    :param path: The path to the file being converted
+    :return: The template string with placeholders replaced
+    """
+
+    return template.format(
+        filepath=path.as_posix(),
+        filename=path.name,
+    )
```
hunyadi: Characters `{` and `}` can occur naturally in text. A less straightforward sequence for [Placeholders](https://docs.gitlab.com/user/markdown/#placeholders) like in GitLab Flavored Markdown would reduce the risk of accidentally marking a piece of text for substitution when no substitution is intended.

tests/test_conversion.py
```diff
@@ -298,6 +298,33 @@ def test_max_image_width_no_constraint(self) -> None:
...
+        "Test that generated_by option throws if it tries to template a not-supported variables"
+        test_file_path = self.source_dir / "basic.md"
+        _, doc = ConfluenceDocument.create(
+            test_file_path,
+            ConfluenceDocumentOptions(generated_by="{template}"),
+            self.source_dir,
+            self.site_metadata,
+            self.page_metadata,
+        )
+        with self.assertRaises(KeyError):
```
hunyadi: `KeyError` is a rather generic exception, which may be hard for library users to distinguish from exceptions originating from elsewhere, I would use a more specific exception class. In particular, *md2conf* defines `ArgumentError`, which is thrown when a configuration parameter has a wrong value, e.g.

```
raise ArgumentError("Confluence API URL or domain required")
```

 I would validate `generated_by` when it is set, e.g. verify that it has only the supported set of placeholders. If something is wrong, we could raise an `ArgumentError`.

md2conf/converter.py
```diff
@@ -1972,6 +1990,7 @@ def __init__(
             generated_by = None
 
         if generated_by is not None:
+            generated_by = apply_generated_by_template(generated_by, path.relative_to(root_dir))
```
hunyadi, line 1996: By itself, this would defer validating the string to the point where the string is first used in a Markdown document. I would recommend implementing an early validation (e.g. when the option is set) to ensure that the rest of the code can assume that `apply_generated_by_template` never fails.

hunyadi (commented):
Thanks for contributing to *md2conf*! I have left some comments to make this feature more robust from an end user perspective.

md2conf/converter.py
baptman21: Agreed, but if I'm not mistaken, it's also possible to set the generated string per document, which means we need to validate twice, no? 
By validating the string, do you mean performing a first template with a placeholder path to check if an ArgumentError occurs?
Do you have pointers for where I should put those checks in the code please? (I'm not that familiar with the codebase yet 😅 )

baptman21: Actually, just pushed a different version with a simple replace logic, this makes it impossible to have an invalid field.
Thoughts?

tests/test_conversion.py
baptman21: Linked to the discussion in the following thread, this is no longer relevant. Closing for the next one.

md2conf/converter.py
hunyadi: My first thought was that when we have a `generated_by` string, we first verify if the string has supported placeholders only. For example, if someone writes

```
Check out this file %{file} in the repo
```

then we can immediately warn them that we don't support the placeholder `%{file}`, even before processing any documents. Verification could happen when the string is assigned to `generated_by` (e.g. in the object initializer). If the body of the function `apply_generated_by_template` is kept along the member variable `generated_by` (in the same class), verifying the value and applying the value would more likely to remain in sync.

I am happy to make this change as a follow up. Your approach of replacing only the supported placeholders (and never triggering a `KeyError`) is also totally fine.

README.md
```diff
@@ -169,6 +169,17 @@ Provide generated-by prompt text in the Markdown file with a tag:
 
 Alternatively, use the `--generated-by GENERATED_BY` option. The tag takes precedence.
 
+The generated-by text can also be templated with the following variables:
+
+- `%{filename}`: the name of the markdown file
+- `%{filepath}`: the path of the markdown file relative to the root
```
hunyadi: The rest of the `README` starts *Markdown* with a capital *M*. It may help readers if we stay consistent.

Even though I completely understand what *root* refers to, `README` mentions *root* only in the context of a *root page* (in Confluence). We might need to use a different phrase, or explain what a *root* is. (For example, what does this mean when you publish a single file?)

md2conf/converter.py
baptman21: Ack, thanks! I'll keep this approach for the PR then for now.

README.md
baptman21: Fixed the Markdown.

I replaced it with `source directory` as it is mentioned in the `--keep-hierarchy` option's help message.

Issue #116, opened 2025-05-27
Document title appears twice on page


florianhofmaier:
The Document title is taken from the first header in the document and assigned as page title.
Confluence shows the title on top of the page. it would be nice when it will be removed from the body, because now it appears twice on the page.

PR #182, opened 2025-12-15
feat: add CLI argument to skip the first heading when used as page title

codemedic:
Implements what is suggested in https://github.com/hunyadi/md2conf/issues/116#issuecomment-2916491336


md2conf/converter.py
```diff
@@ -2013,6 +2015,49 @@ def __init__(
...
+        Removes the first heading element from the document root.
+
+        This is used when the title was extracted from the first unique top-level heading
+        and the user has requested to skip it from the body to avoid duplication.
+
+        Handles the case where a generated-by info panel may be present as the first child.
+        """
+
+        # Find the first heading element (h1-h6) in the root
+        heading_pattern = re.compile(r"^h[1-6]$", re.IGNORECASE)
```
hunyadi, line 2062: One mild risk I see here is when the article has an abstract, and then a top-level heading, e.g.

```markdown

This is an abstract.

# My Title

This is the rest of the document.
```

*md2conf* would extract *My Title*, and then the abstract and the rest of the document would collapse. However, the document structure would still remain reasonable, even if the abstract would become less separated.

md2conf/converter.py
```diff
@@ -2013,6 +2015,49 @@ def __init__(
...
+
+        for child in self.root:
+            if heading_pattern.match(child.tag):
```
hunyadi: The code is probably easier to read with less indentation:

```python
if heading_pattern.match(child.tag) is None:
    continue
```

md2conf/converter.py
```diff
@@ -2013,6 +2015,49 @@ def __init__(
...
+                # Preserve any text that comes after the heading (tail text)
+                tail = child.tail
+
+                # Remove the heading
+                self.root.remove(child)
+
+                # If there was tail text, attach it to the next element
```
hunyadi: The name `next_elem` might be misleading. For example, take the following HTML:

```html
<body><article>A <h1>heading</h1> with <em>a</em> tail</article></body>
```

Then the `tail` of the element `<h1>` (i.e. ` with `) should be appended to the `text` of `<article>`, not of `<em>` or `<body>`.

md2conf/converter.py
codemedic: Should I consider an alternative approach; e.g., only remove heading if there's no content before it?

.. or would you be OK if this behaviour is documented and a unit test added to formalise this behaviour?

hunyadi: I think this is rather an edge case, which is why I wrote *mild risk*. Since the implementation defaults to `--no-skip-title-heading`, and the document maintains a reasonable structure even if the rest of the document immediately follows the abstract, I would say it's OK to keep the behavior as proposed in this pull request. If you add/update a unit test to showcase this behavior, it would be great!

codemedic: Cheers @hunyadi .. I have added a mention of this behaviour in the README and added a test for it.

PR #183, opened 2025-12-16
feat: update Docker workflow to include unstable tag

codemedic:


.github/workflows/publish-docker.yml
```diff
@@ -2,21 +2,32 @@ name: Publish Docker image
 
 on:
   push:
+    branches:
+      - master
```
hunyadi, line 6: I don't think I am planning on publishing Docker images from each commit of the `master` branch. There are a number of maintenance (housekeeping) commits on `master` that should definitely not trigger a Docker build (or PyPI release). For example, contributing authors sometimes miss running the formatter/linter, forget to add unit tests, or inadvertently introduce minor bugs, which are addressed in follow-up commits. When there is activity on the `master` branch, *md2conf* releases are relatively frequent, making an `unstable` tag less critical. You can always clone the repository or instruct `pip` to download and install from the latest source code if needed.

codemedic: I am currently consuming md2conf via the docker image, within a github workflow.
I was hoping I can test / make use of the latest changes in `master` using the `unstable` tag.

If there are frequent tagged releases of md2conf, then it is less of a worry, but would have benefited me to test the latest changes.

hunyadi: Normally, we batch 4-5 fixes/improvements (a "release train") and then publish a patch release (e.g. `0.5.2`). We run the integration test suite before each patch release (in a Confluence Cloud environment), which sometimes flags issues that we need to address. If everything passes, the packages are pushed to PyPI and the public Docker container registry. The next release is planned in the next few days.

codemedic: Thanks @hunyadi

Issue #186, opened 2025-12-19
[Feature] Add PlantUML diagram support


codemedic:
### Describe your idea

Add support for PlantUML diagrams in Markdown documents, enabling users to include UML and other PlantUML diagrams that are either pre-rendered to images or embedded as Confluence macros.

## Motivation

PlantUML is a widely-used tool for creating UML diagrams from plain text, similar to how Mermaid is used for various diagram types. Users working with technical documentation often need to include class diagrams, sequence diagrams, component diagrams, and other UML visualizations that PlantUML excels at generating.

This feature would provide consistency with the existing Mermaid and Draw.io support, giving users flexibility in choosing their preferred diagramming tool.

### Two Rendering Modes

1. **Embedding Mode (Default)**: Inserts Confluence `plantuml` macros that render diagrams within Confluence
2. **Rendering Mode**: Pre-renders diagrams to PNG/SVG images using the PlantUML CLI

### Supported Formats

**Fenced code blocks:**
````markdown
```plantuml
@startuml
class Example {
  +method()
}
@enduml
```
````

**External files:**
- `.puml` files
- `.plantuml` files

### Configuration Options

Per-diagram configuration via YAML frontmatter:
````yaml
---
title: My Class Diagram
config:
    scale: 1.5
---
@startuml
class Example
@enduml
````

### Technical Implementation

The implementation would follow the same architectural patterns as Mermaid and Draw.io support:

- **Scanner** (`scanner.py`): Would discover `.puml` and `.plantuml` files in the document tree
- **Domain** (`domain.py`): Would define `PlantUMLDocument` and `PlantUMLFigure` document types
- **PlantUML Module** (`plantuml.py`):
  - Would handle rendering via PlantUML CLI
  - Would extract and parse YAML frontmatter metadata
  - Would manage configuration properties (scale, etc.)
- **Converter** (`converter.py`):
  - Would generate Confluence Storage Format macros for embedding mode
  - Would render to images for rendering mode
  - Would handle both external files and fenced code blocks

### Dependencies

Requires PlantUML CLI to be installed and available in PATH:
- **Java runtime** (JRE 8 or later)
- **PlantUML JAR file** (can be downloaded from https://plantuml.com/download)

Users can verify their setup with:
```bash
java -jar plantuml.jar -version
```

### Usage Examples

**Using fenced code blocks:**
````markdown
```plantuml
@startuml
Alice -> Bob: Authentication Request
Bob --> Alice: Authentication Response
@enduml
[...]

PR #187, opened 2025-12-19
Add plantuml support

codemedic:
Implements #186


plantuml.sh
hunyadi, line 1: *md2conf* currently strives to be OS-independent: it works on Linux, macOS and Windows. In particular, some of the current *md2conf* users trigger synchronization from a Windows machine. On the other hand, I would refrain from writing OS-specific install scripts for each target OS, which would increase maintenance burden. We might want to consider linking to official installation instructions provided by PlantUML, and assume a working installation is available when *md2conf* runs.

md2conf/plantuml.py
```diff
@@ -0,0 +1,95 @@
...
+            f"failed to convert PlantUML diagram; "
+            f"exit code: {proc.returncode}"
+        ]
+        console_output = stdout.decode("utf-8")
+        if console_output:
+            messages.append(f"output:\n{console_output}")
+        console_error = stderr.decode("utf-8")
+        if console_error:
+            messages.append(f"error:\n{console_error}")
+        raise RuntimeError("\n".join(messages))
```
hunyadi: This looks strikingly similar to how `subprocess` is invoked for Mermaid diagrams (via `mmdc`). If this is shared code, we might want to move this to a separate function or base class.

md2conf/converter.py
```diff
@@ -1104,6 +1113,157 @@ def _create_mermaid_embed(self, filename: str) -> ElementType:
...
+                        width=svg_width,
+                        height=svg_height,
+                        alt=None,
+                        title=None,
+                        caption=None,
+                        alignment=ImageAlignment(self.options.alignment),
+                        display_width=self.options.calculate_display_width(
+                            svg_width
+                        ),
+                    )
```
hunyadi: These steps look similar to those needed for Mermaid diagrams (e.g. post-processing SVG output calls the same functions in the same order). I would look into how we can extract shared functionality into a common function that we call for both Mermaid and PlantUML.

Dockerfile
```diff
@@ -35,13 +35,22 @@ ENV CHROME_BIN="/usr/bin/chromium-browser" \
...
     && apk add chromium font-noto-cjk font-noto-emoji terminus-font ttf-dejavu ttf-freefont ttf-font-awesome ttf-inconsolata ttf-linux-libertine \
     && fc-cache -f
 
 RUN addgroup md2conf && adduser -D -G md2conf md2conf
 USER md2conf
 WORKDIR /home/md2conf
+
+# Copy plantuml.sh script and set it up
+COPY --chown=md2conf:md2conf plantuml.sh /home/md2conf/
+RUN chmod +x /home/md2conf/plantuml.sh \
```
hunyadi: The PlantUML version we install appears to be embedded in `plantuml.sh`. It might make sense to configure the version number at a single place, alongside the version for Mermaid (see top of this file), to help future maintainers as new versions are released.

Dockerfile
```diff
@@ -35,13 +35,22 @@ ENV CHROME_BIN="/usr/bin/chromium-browser" \
 # install dependencies for @mermaid-js/mermaid-cli
 # https://github.com/mermaid-js/mermaid-cli/blob/master/install-dependencies.sh
 RUN apk upgrade \
-    && apk add --update nodejs npm \
+    && apk add --update nodejs npm curl openjdk17-jre-headless graphviz \
```
hunyadi: Mermaid has already forced us to package a Node.js environment such that `mmdc` can run. PlantUML adds a full JRE as a requirement. The Docker image for *md2conf* is already >600 MB. We might want to re-assess how we distribute a self-contained *md2conf* given that not everyone might want to use Mermaid and/or PlantUML rendering features.

plantuml.sh
codemedic: Thanks for the feedbac @hunyadi.

Perhaps the best option is, as you suggested, linking the official documentation for PlantUML. In combination with that, we could also invoke the jar file directly from python (using JRE). The command construct for that would be fairly cross-platform, assuming we can getaway with invoking `java` as just `java` instead of `java.exe` on Windows.

md2conf/plantuml.py
codemedic: I will give that a go.

md2conf/converter.py
codemedic: I am fairly new to python but I will give that a go.

Dockerfile
codemedic: I had similar concerns while making these changes. An alternative I can think of is to use multi-stage build to generate multiple images (base, base+mmd, base+plantuml, base+all).

codemedic: I have implemented a multi-stage build based approach.
Please let me know what you think.

hunyadi: Looks great! Initially, I had thought of having a bare-bone version (no Mermaid or PlantUML) and a full version (both Mermaid and PlantUML). However, it definitely makes sense to enable them one by one, especially given that PlantUML is somewhat lighter than Mermaid. I am no Docker or GitHub Actions wizard but the overall structure seems OK.

codemedic: Appreciate the feedback @hunyadi.
I will submit the finalised PR after I have done some more testing.

codemedic: I have tested the build.

Please see the workflow run results
https://github.com/codemedic/hunyadi-md2conf/actions/runs/20378162210

Build results can be found in docker hub.
https://hub.docker.com/r/codemedic/md2conf/tags

hunyadi:
> Do you have the testing process documented somewhere?

`CONTRIBUTING.md` has a short section about *Running integration tests*. The folder `integration_tests` contains those tests that require a real Confluence environment, and synchronize Markdown documents in the folder `sample`. We have a Confluence Cloud account set up, and we pass the domain, user name and API keys in environment variables. Today, `integration_tests` has some "hidden assumptions" about the existence of some Confluence pages, these are declared in the preamble of `test_api.py` and `test_csf.py`, e.g. `FEATURE_TEST_PAGE_ID` or `IMAGE_TEST_PAGE_ID`. If you wanted to use your own environment, you would have to create these pages, and set their page ID in the preamble, replacing existing values. It would have been nice if we made it possible to run `integation_tests` from scratch without any preliminary configuration but this has unfortunately not received too much attention yet. (Required API calls are implemented in `api.py` but tests in `integration_tests` are not wired to check for the presence of these pages and create them dynamically if missing.) There is also a human element involved, e.g. we would manually verify if PlantUML `structured-macro` has been properly generated (from `sample/attachments.md`) by visiting the corresponding Confluence page once `integration_tests` is finished running.

README.md
```diff
@@ -579,15 +600,17 @@ positional arguments:
...
+  -u USERNAME, --username USERNAME
                         Confluence user name.
-  -a, --api-key API_KEY
+  -a API_KEY, --api-key API_KEY
                         Confluence API key. Refer to documentation how to obtain one.
-  -s, --space SPACE     Confluence space key for pages to be published. If omitted, will default to user space.
-  -l, --loglevel {debug,info,warning,error,critical}
+  -s SPACE, --space SPACE
+                        Confluence space key for pages to be published. If omitted, will default to user space.
+  -l {debug,info,warning,error,critical}, --loglevel {debug,info,warning,error,critical}
```
hunyadi: These don't appear to be real changes. Python 3.13 displays the help message is a more compact format, e.g. `-u, --username USERNAME`. Earlier versions of Python use a more verbose format, e.g. `-u USERNAME, --username USERNAME`. I prefer the format with less repetition.

integration_tests/README.md
```diff
@@ -0,0 +1,372 @@
...
+When you run tests for the first time:
+
+```
+setUpModule()
+  ├─ Read CONFLUENCE_INTEGRATION_TEST_PARENT_PAGE_ID from environment
+  ├─ Create "Publish Markdown to Confluence" page → ID: 917515
+  ├─ Create "Test Page for Attachments" page → ID: 917534
+  └─ _write_page_ids_to_samples()
+       ├─ Create/find pages for each sample file
+       └─ Inject page IDs into markdown files:
```
hunyadi: Injecting IDs is not strictly necessary; *md2conf* is able to look up pages based on their title if no ID is present. Early versions of *md2conf* (when many integration tests were originally written) did not have this capability. It might be possible to rewrite some of the tests to use title lookup instead of ID lookup. I am happy to look into this.

integration_tests/README.md
```diff
@@ -0,0 +1,372 @@
...
+
+**Expected Behavior**: Sample files get `<!-- confluence-page-id: ... -->` comments injected
+
+**Note**: This is **intentional** - it makes tests realistic and self-contained. These IDs are specific to your test environment and should not be committed.
+
+## Design Decisions
+
+### Why Inject Page IDs into Sample Files?
+
+1. **Realistic examples**: Sample files show actual usage with page IDs
```
hunyadi: While this is true, it would probably suffice to have one or two files that are updated based on ID (to cover this feature), and synchronize the rest based on title matching (which simplifies writing integration tests). Today, many of the sample files have ID-based synchronization purely for historical reasons. I am glad to look into this in more depth.

integration_tests/README.md
```diff
@@ -0,0 +1,372 @@
...
+3. **Cleanup control**: Optional deletion via environment variable
+
+## Contributing
+
+When adding new integration tests:
+
+1. **Add sample file** to `../sample/` if needed
+2. **Update `sample_files` dict** in `_write_page_ids_to_samples()`
+3. **Use existing test pages** where possible (reuse `FEATURE_TEST_PAGE_ID`)
+4. **Document new tests** in this README
```
hunyadi: Based on my experience, this is unlikely to happen. Unless we have explicit coverage for documentation, it is likely to quickly go out of sync with the code, as future maintainers make changes to code (and tests) but simply forget to update any documentation they don't get notified that they should update. This is why I would prefer tests to be self-contained and self-explanatory. We can have a general description for integration tests overall (which is great) but I would put test-specific documentation in the tests themselves; I would not maintain any other "catalog" of tests.

integration_tests/fixtures.py
```diff
@@ -0,0 +1,230 @@
...
+        >>>     page_id = fixture.get_or_create_test_page("My Test Page")
+        >>>     # Run tests with page_id
+        >>>     fixture.cleanup(delete_pages=False)
+    """
+
+    def __init__(
+        self,
+        session: ConfluenceSession,
+        space_key: str,
+        cache_file: Optional[Path] = None,
```
hunyadi: Minor remark: the rest of the project uses the union syntax `Path | None` to denote optional values.

integration_tests/test_api.py
```diff
@@ -16,23 +17,154 @@
...
+    sample_files = {
+        "index.md": "Publish Markdown to Confluence",
+        "code.md": "Fenced code blocks",
+        "attachments.md": "Images and documents",
+        "panel.md": "Admonitions and alerts",
+        "plantuml.md": "PlantUML Diagrams",
+        "sibling.md": "Markdown example document",
+        "parent/index.md": "🏠 Markdown parent page",
+        "parent/child.md": "Markdown child page",
+    }
```
hunyadi: `SynchronizingProcessor` can sync entire directory hierarchies (not only single files) so we can point it to the directory `sample`. This would be more robust as future maintainers add/remove files to `sample`.

integration_tests/fixtures.py
hunyadi, line 1: Following the most recent commit to `master`, this utility module should no longer be necessary. All hard-wired identifiers have been removed from sample Markdown files and integration tests. Tests are now creating dependent Confluence pages in the `setUpClass` class method of the test suite once per test suite instantiation. The implementation in `setUpClass` looks up pages by their title, and if not found, they are created as child pages of the Confluence space homepage.

integration_tests/README.md
codemedic: The new approach makes more sense and this README would need a rewrite. So I have removed it for now.

README.md
codemedic: Done

md2conf/converter.py
```diff
@@ -1103,6 +1140,155 @@ def _create_mermaid_embed(self, filename: str) -> ElementType:
...
+            except Exception as e:
+                LOGGER.warning(f"Failed to render PlantUML diagram from {absolute_path}: {e}. Embedding without dimensions.")
+        elif self.options.render_plantuml and not plantuml.has_plantuml():
+            LOGGER.warning("PlantUML JAR not available; diagrams will display without dimensions. See README for installation.")
+
+        # Track original file as image for reference
+        self.images.append(ImageData(absolute_path, attrs.alt))
+
+        # Always create plantumlcloud macro with compressed source
+        return self._create_plantuml_macro(svg_filename, content, width, height)
```
hunyadi: Can you double-check this implementation for the use case when `render_plantuml` is enabled? *md2conf* help text says this option *Render[s] PlantUML diagrams as image files*. However, even if I have `render_plantuml` enabled, the code outputs a `structured-macro` with a `name` of `plantumlcloud`. Since I don't have the PlantUML Marketplace extension installed at the moment, the Confluence page says: *Error loading the extension!*

If I compare the implementation to that of *Mermaid*, I am missing a call to `self._create_attached_image` when `render_plantuml` is enabled.

md2conf/converter.py
```diff
@@ -1103,6 +1140,155 @@ def _create_mermaid_embed(self, filename: str) -> ElementType:
...
+
+                # Add SVG as attachment
+                self.embedded_files[svg_filename] = EmbeddedFileData(image_data)
+            except Exception as e:
+                LOGGER.warning(f"Failed to render PlantUML diagram: {e}. Embedding without dimensions.")
+        elif self.options.render_plantuml and not plantuml.has_plantuml():
+            LOGGER.warning("PlantUML JAR not available; diagrams will display without dimensions. See README for installation.")
+
+        # Always create plantumlcloud macro with compressed source
+        return self._create_plantuml_macro(svg_filename, content, width, height)
```
hunyadi: If I compare the implementation to that of *Mermaid*, I am missing a call to `self._create_attached_image` and an early return when `render_plantuml` is enabled. Currently, it seems that we always generate a `structured-macro` even if the user is explicitly asking for a PNG/SVG image file to be generated and uploaded as an attachment.

codemedic: Sorry about that.

I will take a look in a few hours; I have a busy day today.

codemedic: The issues must be sorted now

Also, I have added additional inputs to the github action workflow for integration tests to control rendering of mermaid and plantuml, which allows easier control of behaviour.

hunyadi: Great! I have made minor changes to unit and manual tests to make sure they pass. `render_plantuml=True` now works as expected. I will install the PlantUML Marketplace extension later today to verify it also works as planned.

Issue #189, opened 2025-12-23
[Feature] Add PlantUML include path and theme configuration options


codemedic:
## Problem

When publishing PlantUML diagrams to Confluence, users often want to:
1. Use `!include` directives to reference shared theme files and common definitions
2. Apply consistent themes across all diagrams
3. Pre-include theme loader files that reference remote themes

Currently, these native PlantUML features don't work properly because:
- PlantUML's `!include` directive cannot resolve relative paths from the project root
- No way to apply a global theme without modifying each diagram
- No way to pre-include files before diagram processing

## Proposed Solution

Add three new CLI options to configure PlantUML rendering:

```bash
--plantuml-include-path PLANTUML_INCLUDE_PATH
    Include path for resolving PlantUML !include directives (default: current working directory)

--plantuml-include PLANTUML_INCLUDE
    File to pre-include before processing PlantUML diagrams (relative to include path)

--plantuml-theme PLANTUML_THEME
    Built-in PlantUML theme name to apply to all diagrams
```

## Use Cases

**1. Shared theme definitions:**
```bash
md2conf --plantuml-include-path /path/to/project \
        --plantuml-include themes/common.puml \
        docs/
```

**2. Remote theme loading:**
Create `themes/loader.puml`:
```plantuml
!theme mytheme from https://raw.githubusercontent.com/user/repo/main/themes
```

Then use:
```bash
md2conf --plantuml-include-path /path/to/project \
        --plantuml-include themes/loader.puml \
        docs/
```

**3. Built-in theme:**
```bash
md2conf --plantuml-theme cerulean docs/
```

## Benefits

- Enables native PlantUML features without workarounds
- Consistent theming across all diagrams
- Supports both local and remote theme repositories

hunyadi:
Looking at the implementation proposal #190, command-line options `--plantuml-include-path`, `--plantuml-include` and `--plantuml-theme` only appear to impact how the external PlantUML executable is invoked, not how Confluence Storage Format XHTML is generated for the third-party Marketplace extension. As such, a seemingly much more straightforward approach would be to configure the shell command for the PlantUML executable. Today, it is roughly as follows:

```sh
$ java -jar plantuml.jar -pipe -scale 2
```

However, we could easily update the command (e.g. by means of an environment variable) to read as follows (with placeholders for substituted values):

```sh
[...]

Issue #188, opened 2025-12-19
[Feature] Discussion: Configurable PlantUML macro name and parameters


codemedic:
### Describe your idea

Right now, the PlantUML macro name (`"plantuml"`) and its parameters are hardcoded. While this might work fine for standard Confluence Cloud setups, it creates friction for teams with custom configurations:

- **Third-party apps**: Organizations often install alternative PlantUML apps from the Atlassian Marketplace that use different macro names (e.g., `plantuml-cloud`, `com.github.confluence.plantuml`)
- **Custom parameters**: Users may need to configure output format, themes, server URLs, or caching options that aren't exposed with the current hardcoded approach
- **Advanced features**: Some PlantUML apps support additional parameters for performance tuning or custom styling

### Current Implementation

Here's what the code looks like today:

```python
def _create_plantuml_embed(self, filename: str) -> ElementType:
    return AC_ELEM(
        "structured-macro",
        {AC_ATTR("name"): "plantuml", ...},  # ← HARDCODED
        AC_ELEM("parameter", 
            {AC_ATTR("name"): "atlassian-macro-output-type"}, 
            "INLINE"  # ← HARDCODED
        ),
        AC_ELEM("plain-text-body", ET.CDATA(filename)),
    )
```

For context, Mermaid and Draw.io also use hardcoded macro names—this works for most users since those macros are fairly standardized across Confluence Cloud installations. However, PlantUML is a different story.

## Proposed Approach

What if we made the macro name and parameters configurable through CLI arguments, per-diagram frontmatter, and optionally a global configuration file? This would give users the flexibility they need without overcomplicating the common case.

### Command-Line Configuration

```bash
md2conf \
  --plantuml-macro-name "plantuml-cloud" \
  --plantuml-macro-params '{"output":"svg","theme":"dark"}'
```

### Per-Diagram Configuration

````yaml
---
title: My Sequence Diagram
config:
    scale: 1.5
    macro_name: "plantuml-cloud"
    macro_params:
        output: "svg"
        theme: "dark"
---
@startuml
Alice -> Bob: Hello
@enduml
````

### Global Configuration File (Optional)

This could be a nice-to-have that benefits all features, not just PlantUML:

```yaml
# .md2conf.yaml
confluence:
  domain: example.atlassian.net
  space: DOCS

rendering:
  mermaid: true
  drawio: false
  plantuml: false

plantuml:
  macro_name: "plantuml"
  macro_params:
    output: "svg"
```

[...]

PR #190, opened 2025-12-23
feat: add PlantUML include path and theme support

codemedic:
#189 - Add CLI options to configure PlantUML rendering:
- `--plantuml-include-path`: sets Java property for resolving !include directives
- `--plantuml-include`: pre-includes a file (e.g., theme loader) before processing
- `--plantuml-theme`: applies built-in theme to all diagrams

Enables native PlantUML features (!include, !theme) to work with relative paths. Per-diagram scale can still be set via YAML front-matter.

### DIFF
--- a/md2conf/plantuml.py
+++ b/md2conf/plantuml.py
@@ -99,10 +99,8 @@ def get_plantuml_command(include_path: str | None = None) -> list[str]:
 
     # JAR not found - fail with helpful message
     raise RuntimeError(
-        f"PlantUML JAR not found. Download `plantuml.jar` from "
-        f"https://github.com/plantuml/plantuml/releases "
-        f"and place it at {jar_path}, or set the PLANTUML_JAR "
-        f"environment variable to point to it."
+        f"PlantUML JAR not found. Download `plantuml.jar` from https://github.com/plantuml/plantuml/releases "
+        f"and place it at {jar_path}, or set the PLANTUML_JAR environment variable to point to it."
     )
 
 


integration_tests/test_api.py
```diff
@@ -31,6 +31,17 @@
...
+# Configure logging for integration tests
+# Use environment variable to control level (default to INFO)
+_log_level_str = os.getenv("LOG_LEVEL", "INFO").upper()
+_log_level = getattr(logging, _log_level_str, logging.INFO)
+
+logging.basicConfig(
+    level=_log_level,
+    format=("%(asctime)s - %(levelname)s - %(name)s - %(funcName)s [%(lineno)d] - %(message)s"),
+    force=True,  # Override any existing configuration
+)
```
hunyadi, line 43: This file already configures logging in its `if __name__ == "__main__":` block, see below. I am afraid I don't agree with adjusting logging properties at the module level, as this would overwrite logging configuration for whichever other module that happens to import any symbols from this module. In my opinion, a better place to configure logging properties is the entry point for the runner. This would decouple logging configuration from the tests themselves.

md2conf/__main__.py
```diff
@@ -205,6 +208,24 @@ def get_parser() -> argparse.ArgumentParser:
...
+        dest="plantuml_include",
+        default=None,
+        help="File to pre-include before processing PlantUML diagrams (relative to include path).",
+    )
+    parser.add_argument(
+        "--plantuml-theme",
+        dest="plantuml_theme",
+        default=None,
+        help="Built-in PlantUML theme name to apply to all diagrams.",
+    )
```
hunyadi, line 228: I am not sure how comprehensive this set of options is (e.g. how many new PlantUML options we may want to add in the future). If this is rather open-ended, it might be better to propagate these with another configuration mechanism, e.g. a JSON or YAML file, which we can directly parse into Python objects with `cattrs`. I would like to avoid overwhelming the user with a ton of options they might never use, e.g. these options don't add value to someone who uses Mermaid only.

md2conf/__main__.py
```diff
@@ -348,6 +369,9 @@ def main() -> None:
         render_drawio=args.render_drawio,
         render_mermaid=args.render_mermaid,
         render_plantuml=args.render_plantuml,
+        plantuml_include_path=args.plantuml_include_path,
+        plantuml_include=args.plantuml_include,
+        plantuml_theme=args.plantuml_theme,
```
hunyadi, line 374: I would suggest encapsulating all PlantUML options in a single `PlantUMLOptions` object, see `LayoutOptions` below for reference. This would also make it easier to refactor the project once we move to a plugin system where add-ons like Mermaid or PlantUML are implemented in their separate modules that get access to the context (e.g. relative path of file they are operating in), interact with the context (e.g. add a new image to the list of attachments to synchronize), and ultimately emit Confluence Storage Format XHTML.

md2conf/domain.py
```diff
@@ -70,6 +73,9 @@ class ConfluenceDocumentOptions:
     render_drawio: bool = False
     render_mermaid: bool = False
     render_plantuml: bool = False
+    plantuml_include_path: str | None = None
+    plantuml_include: str | None = None
+    plantuml_theme: str | None = None
```
hunyadi, line 78: Some of these repetitions can be avoided if these are fields of their own class, see `LayoutOptions` below.

md2conf/__main__.py
codemedic: I couldn't agree more. My intention was to implement the bare minimum options I needed for my use-case while keeping them generic.

It might make better sense to implement these changes based on the proposal here https://github.com/hunyadi/md2conf/issues/188#issuecomment-3689574787

hunyadi: If you are comfortable moving forward with these limited set of PlantUML options, I am happy to review the proposal as it is. However, as you can see, a lot of boilerplate is involved propagating command-line options to the converter class, which is why I would like to move towards a plugin architecture where plugins can be improved independently of the core, and plugin configuration can completely bypass the Markdown converter. On the other hand, I will likely have less time than usual to devote to projects during the festive season, and I may need to ask for more patience in getting back to you.

As a compromise, you might want to take a look how `LayoutOptions` is implemented, which reduces the amount of boilerplate needed, even if it still follows the same propagation path as regular options do. This might be a trade-off solution until the foundations for the plugin architecture are laid. The compromise solution would substantially reduce the amount of code we need to migrate when we make the switch.

PR #191, opened 2025-12-30
Add GitHub action

codemedic:
### New GitHub Action
I’ve introduced a **composite GitHub Action** that runs `md2conf` within a Docker container. This simplifies CI/CD integration by removing the need for users to manage Python environments or dependencies manually.

* **Documentation:** Added `ACTION.md` with implementation examples.
* **Portable:** The Docker-based approach ensures consistent behavior across different CI runners.

### Release Automation & v1.0.0 Transition
To support the Action and modernize our release flow, I’m proposing a shift to `v{major}.{minor}.{patch}` starting with **v1.0.0**:

* **Major Version Tracking:** Transitioning to `1.0.0` allows us to support floating tags (e.g., `@v1`). This is the standard for GitHub Actions, letting users receive non-breaking updates without pinning to specific patches.
* **The `v` Prefix:** Switching from `0.5.2` to `v1.0.0` aligns with GitHub's release conventions and simplifies automated tag management.
* **Single Source of Truth:** Integrated `setuptools_scm` to derive package versions directly from Git tags. This ensures PyPI metadata is always in sync and eliminates the need for manual version bumps in the source code.

---

I'm interested in your thoughts on the `v` prefix and the jump to `1.0.0`. While it's a change in convention, it seems like the most robust way to support the new Action ecosystem.

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -55,7 +55,7 @@ jobs:
         with:
           images: ${{ env.DOCKER_IMAGE_NAME }}
           tags: |
-            # For tags: create versioned tags 
+            # For tags: create versioned tags
             type=semver,pattern={{version}}${{ matrix.suffix }}
             type=semver,pattern={{major}}${{ matrix.suffix }}
             # For tags: create latest tag


codemedic:
@hunyadi I have some more optimisations that would benefit both docker image publishing and GitHub action testing workflows.

This PR is already too long so I am reluctant to add more to it, unless you are happy to look at them together.

For now, I have created a draft PR in my fork. Please take a look and let me know.
https://github.com/codemedic/hunyadi-md2conf/pull/1

PR #203, opened 2026-01-12
feat: improve documentation help formatting in README.md

codemedic:
This PR improves the formatting of the help section in `README.md` and automates previous manual maintenance tasks.

Key changes:
- Using a custom help formatter to prevent line wrapping (increased width).
- Automatically removing redundant metavariables for options with short flags (e.g., `-u USERNAME, --username USERNAME` -> `-u, --username USERNAME`). Previously, contributors had to manually remove these redundant metavariables from the documentation.
- Exposing `formatter_class` in `get_parser` and `get_help` to allow this customisation.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -131,6 +131,9 @@ gh run watch <run-id> --repo <owner>/<repo>
 
 Verify that all code you have added passes static code checks. Depending on your OS, there is a script for you to use. If using Windows, run `check.bat`, otherwise run `./check.sh`.
 
+> [!NOTE]
+> Documentation generation in the check scripts is skipped if the Python version is older than 3.13 to maintain consistent formatting. You can still run `python documentation.py` manually if needed.
+
 ### Generating documentation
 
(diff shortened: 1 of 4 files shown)


hunyadi:
I understand the intent but I admit that I never touch the help text listing of *md2conf* by hand. I use Python 3.13, which already emits the compact `-u, --username USERNAME` format out of the box, and my console screen width is set to 160 characters to avoid line breaks, which is accomplished on Windows with the following command:

```bat
mode con: cols=160
```

The help text formatter inherits the console screen width, so it produces the right output. If I run `python documentation.py` with these settings, it keeps the formatting of `README.md` as seen in the repository.

codemedic:
Latest updates should cover all the points we discussed!
I'll leave it in your hands now; feel free to merge or close as you see fit.
No worries either way.

hunyadi:
Unfortunately, I still don't like relying on undocumented private variables like `width` or private methods like `_get_default_metavar_for_optional`. On the other hand, I wanted to take your problem seriously since you have already contributed a lot to the project. (Thank you!) As a "best of both worlds" solution, I made some changes to `documentation.py` such that verbose options are replaced with their compact equivalent and spacing is re-aligned as necessary. Tested with both Python 3.10 and 3.13, the output is now identical. (Python 3.13 makes no replacements in the help text.) Changes don't impact anything that is part of core *md2conf*, minimizing risk.

Issue #202, opened 2026-01-10
[Feature] Automate Docker Hub repository description


codemedic:
### Describe your idea

Building on the Docker build summary work in PR #198, this proposal is to automate the update of the Docker Hub repository overview.

## Summary
- Keep Docker Hub documentation in sync with the repository using a managed template.
- Use the existing build workflow to render and publish the description automatically upon release.
- Ensure the publishing step is resilient to metadata permission variations in CI/CD environments.
- Document the necessary authentication setup for maintainers to enable this automation.

## Rationale
Manual updates to the Docker Hub description are prone to becoming outdated. Automating this process ensures that information about variants, tags, and usage remains accurate after every release.

## Dependencies
- Dependent on PR #198.

PR #207, opened 2026-01-14
Update docker hub image description

codemedic:
#202 

This pull request introduces automated generation and validation of the Docker Hub repository description, improving workflow reliability and documentation accuracy. The changes add a new `DOCKER_HUB.md` template, a Python script to generate and validate the description, and update the GitHub Actions workflow to support updating the Docker Hub README during both releases and manual workflow dispatches. Documentation is also updated to reflect these improvements and clarify Docker Hub credential requirements.

### DIFF
--- a/util/dockerhub_description.py
+++ b/util/dockerhub_description.py
@@ -70,9 +70,7 @@ def validate_sync() -> None:
             missing_placeholders.append(placeholder)
 
     if missing_placeholders:
-        error(f"The following placeholders are missing from {TEMPLATE_FILE}:")
-        for p in missing_placeholders:
-            print(f"  - %{{{p}}}", file=sys.stderr)
+        error(f"The following placeholders are missing from {TEMPLATE_FILE}:\n" + "\n".join(f"  - %{{{p}}}" for p in missing_placeholders))
         sys.exit(1)
 
     print(f"Success: {TEMPLATE_FILE} is in sync with {BAKE_FILE}.")


.github/workflows/publish-docker.yml
```diff
@@ -13,6 +13,11 @@ on:
         required: false
         default: true
         type: boolean
+      update_description:
+        description: 'Update DOCKER_HUB.md description'
+        required: false
+        default: false
+        type: boolean
```
hunyadi, line 20: Shouldn't we always update the Docker description? Specifically, is there a use case when we shouldn't? I would think this should happen when you push a tag, and not otherwise.

.github/workflows/publish-docker.yml
```diff
@@ -97,3 +102,47 @@ jobs:
...
+          
+          # Prepare tag lists for each variant (strip image repository name)
+          JSON_TAGS='${{ steps.meta.outputs.json }}'
+          BASE_TAGS=$(echo "$JSON_TAGS" | jq -r '.tags | map(split(":") | last) | map("`" + . + "-minimal`") | join(", ")')
+          MERMAID_TAGS=$(echo "$JSON_TAGS" | jq -r '.tags | map(split(":") | last) | map("`" + . + "-mermaid`") | join(", ")')
+          PLANTUML_TAGS=$(echo "$JSON_TAGS" | jq -r '.tags | map(split(":") | last) | map("`" + . + "-plantuml`") | join(", ")')
+          ALL_TAGS=$(echo "$JSON_TAGS" | jq -r '.tags | map(split(":") | last) | map("`" + . + "`") | join(", ")')
+
+          # Generate final DOCKER_HUB.md using the python script
+          python3 dockerhub_description.py \
```
hunyadi: We might want to put these utilities in a subdirectory called `util` to avoid polluting the root folder. I think we should move `documentation.py` into `util` as well. I have a divided opinion on `check.sh` and some others, which make sense for a developer to execute in the repository root. But `dockerhub_description.py` seems to be more "GitHub Actions" only.

DOCKER_HUB.md
```diff
@@ -0,0 +1,86 @@
+# md2conf - Markdown to Confluence
+
+[Browse code at %{GIT_TAG}](%{GITHUB_REPO_URL}/tree/%{GIT_TAG})
```
hunyadi, line 3: I would recommend we always link to the repository root (base URL). Given our constrained capacity for *md2conf*, we have a policy that only the latest version is supported.

DOCKER_HUB.md
hunyadi, line 1: Do I understand correctly that this file would appear as *Repository overview* on the Docker Hub [project landing page](https://hub.docker.com/repository/docker/leventehunyadi/md2conf/general)?

dockerhub_description.py
```diff
@@ -0,0 +1,126 @@
...
+
+# Mapping between docker-bake.hcl targets and DOCKER_HUB.md placeholders
+TARGET_MAPPING = {
+    "base": "TAGS_BASE",
+    "mermaid": "TAGS_MERMAID",
+    "plantuml": "TAGS_PLANTUML",
+    "all": "TAGS_ALL",
+}
+
+TEMPLATE_FILE = "DOCKER_HUB.md"
```
hunyadi, line 15: We should probably call this file `DOCKER_HUB.md.template` (or something similar) to avoid Markdown rendering in environments like GitHub. After all, it's a partial file with placeholders, not a ready-to-read Markdown document.

dockerhub_description.py
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env python3
+import argparse
+import os
+import re
+import sys
+from typing import Set
```
hunyadi: This is no longer needed in Python 3.9 and later. Since our minimum requirement is Python 3.10 or later. we should use the modern syntax of `set[str]`.

dockerhub_description.py
```diff
@@ -0,0 +1,126 @@
...
+            placeholders.add(m)
+    return placeholders
+
+
+def validate_sync() -> None:
+    """Check if the template is in sync with the bake file."""
+    targets = get_bake_targets(BAKE_FILE)
+    placeholders = get_template_placeholders(TEMPLATE_FILE)
+
+    missing_placeholders = []
```
hunyadi: This should be as below or Pyright will complain:

```python
missing_placeholders: list[str] = []
```

dockerhub_description.py
```diff
@@ -0,0 +1,126 @@
...
+    replacements = {
+        "GIT_TAG": args.git_tag,
+        "DOCKER_IMAGE_NAME": args.image_name,
+        "GITHUB_REPOSITORY": args.github_repo,
+        "GITHUB_REPO_URL": args.github_repo_url,
+        "TAGS_BASE": args.tags_base,
+        "TAGS_MERMAID": args.tags_mermaid,
+        "TAGS_PLANTUML": args.tags_plantuml,
+        "TAGS_ALL": args.tags_all,
+    }
```
hunyadi, line 111: I am a big fan of the static type checker (which catches a ton of coding mistakes) so I would first define a typed namespace of all possible command-line options:

```py
class Arguments(argparse.Namespace):
    check: bool
    git_tag: str | None
    image_name: str | None
    github_repo: str | None
    github_repo_url: str | None
    tags_base: str | None
    tags_mermaid: str | None
    tags_plantuml: str | None
    tags_all: str | None
    output: str
```

Then, parse arguments like as below:

```py
args = Arguments()
parser.parse_args(namespace=args)
```

This will make sure the static type checker has visibility of `args` and check whether all fields are used correctly.

We already do this in `md2conf.__main__` and consistency would help future maintainers.

dockerhub_description.py
codemedic: Done, added the type hint.

dockerhub_description.py
codemedic: Done, implemented the typed `Arguments` class as suggested.

DOCKER_HUB.md
codemedic: Yes, that is correct; it will update the landing page on Docker Hub.

https://hub.docker.com/r/leventehunyadi/md2conf/

(The link you shared is the "private view")

DOCKER_HUB.md
codemedic: Since we always update the landing page for the docker image repo on Docker Hub (within the docker image workflow), this link will always be pointing to the latest tagged version rather than the latest untagged (i.e. possibly unreleased) master branch. This ensures that users landing on the Docker Hub page are directed to stable, released documentation rather than work-in-progress code.

.github/workflows/publish-docker.yml
codemedic: Done, moved `dockerhub_description.py` to the `util/` directory and updated all references.

.github/workflows/publish-docker.yml
codemedic: The `update_description` toggle is a bit of a security fallback. It lets contributors (or folks on forks) manually test the build process without needing to provide credentials or "delete" permissions for a PAT if they aren't worried about the description.

Crucially, the logic on [line 107](https://github.com/codemedic/hunyadi-md2conf/blob/a7151a630e05cffe7711f6907e867a0d45f94158/.github/workflows/publish-docker.yml#L107) ensures that tagged builds always trigger the update. This way, we keep the primary repo's Docker Hub page consistent on release without forcing a strict credential requirement for every manual run.

dockerhub_description.py
codemedic: I went back and forth on this one too! I ended up choosing a `.md` extension so contributors can preview their changes directly in an IDE or on GitHub before submitting. 

That said, since it is a template, would you prefer renaming it to `DOCKER_HUB_TEMPLATE.md` to make that distinction clearer?

dockerhub_description.py
codemedic: Done, switched to modern `set[str]` syntax.

util/dockerhub_description.py
```diff
@@ -0,0 +1,139 @@
...
+    if missing_placeholders:
+        print(f"Error: The following placeholders are missing from {TEMPLATE_FILE}:")
+        for p in missing_placeholders:
+            print(f"  - %{{{p}}}")
+        sys.exit(1)
+
+    print(f"Success: {TEMPLATE_FILE} is in sync with {BAKE_FILE}.")
+
+
+def generate_description(args: argparse.Namespace) -> None:
```
hunyadi: The function parameter needs to be of type `Arguments` (the type-safe derived class), not `argparse.Namespace` (the base class) for the static type checker to have visibility of the fields.

util/dockerhub_description.py
```diff
@@ -0,0 +1,139 @@
...
+    """Generate the final description by replacing placeholders."""
+    if not os.path.exists(TEMPLATE_FILE):
+        print(f"Error: {TEMPLATE_FILE} not found.")
```
hunyadi: This is not a big issue since this is a small utility but standard OS utility practice is to write regular output to `stdout` (normal Python `print`) and errors to `stderr`.

util/dockerhub_description.py
codemedic: Done, updated the function signature to use the typed `Arguments` class.

util/dockerhub_description.py
codemedic: Done, introduced `error()` and `warn()` helpers to DRY up the `stderr` output.

PR #211, opened 2026-01-16
Add the ability to filter out content in original markdown file to prevent that content being sync with Confluence

### LINKED ISSUE #210: [Feature] Support for excluding portions of markdown content from Confluence synchronization
## Issue Statement
When maintaining documentation in both GitLab/GitHub and Confluence, there are cases where certain content should only appear in one platform but not the other. For example:
- References or links that are relevant for repository readers but not for Confluence users
- GitLab/GitHub-specific metadata or navigation elements
- Internal development notes that shouldn't be published to wider audiences on Confluence
- ...

I did not find if md2conf provides such exclusion capabilities.

## Use Case Example and Proposed solution
I have markdown files with references that render properly in GitLab but aren't relevant to Confluence readers:
```md
# Project Documentation

This is the main content that should go to both GitLab and Confluence.

<!-- md2conf-skip-start -->
## Internal References
- See [internal design doc](../internal/design.md)
- Related to issue #123
- For development setup, check the [dev guide](../dev/setup.md)
<!-- md2conf-skip-end -->

## Getting Started
This section should be published to Confluence as meaningful for the end users/readers
```

## Proposed Solution
Add a HTML comment markers to designate sections that should be excluded from Confluence synchronization like shown above

rpioger:
fixes #210

### DIFF
--- a/tests/target/skip_nodes.xml
+++ b/tests/target/skip_nodes.xml
@@ -28,7 +28,6 @@
 	<ac:parameter ac:name="language">py</ac:parameter>
 	<ac:plain-text-body><![CDATA[def foo():
     print("Using skip markers in code block is")
-    <!-- confluence-skip-start -->
-    print("ignored")
-    <!-- confluence-skip-end -->]]></ac:plain-text-body>
+    # <!-- confluence-skip-start -->
+    print("ignored")]]></ac:plain-text-body>
 </ac:structured-macro>


md2conf/text.py
```diff
@@ -52,3 +56,34 @@ def wrap_text(text: str, line_length: int = 160) -> str:
...
+    Removes sections marked with md2conf-skip comments.
+    
+    Removes content between:
+    <!-- md2conf-skip-start --> and <!-- md2conf-skip-end -->
+    
+    :param text: Raw Markdown text
+    :returns: Markdown text with excluded sections removed
+    """
+    # Pattern to match excluded sections
+    # (?s) is inline flag for re.DOTALL - makes . match newlines
```
hunyadi: This looks like a stale comment; you don't seem to be using this inline flag.

md2conf/text.py
```diff
@@ -52,3 +56,34 @@ def wrap_text(text: str, line_length: int = 160) -> str:
...
+    end_pattern = r'<!--\s*md2conf-skip-end\s*-->'
+    start_count = len(re.findall(start_pattern, text, flags=re.IGNORECASE))
+    end_count = len(re.findall(end_pattern, text, flags=re.IGNORECASE))
+    if start_count != end_count:
+        logger.warning(
+            f"Unmatched md2conf-skip markers: found {start_count} start marker(s) "
+            f"and {end_count} end marker(s). Content may not be excluded as expected."
+        )
+    
+    cleaned_text = re.sub(pattern, '', text, flags=re.DOTALL | re.IGNORECASE)
```
hunyadi: What is the use case for `re.IGNORECASE`? I think the exclusion pattern has to be as strict as reasonably possible to minimize the chances of unintended matches.

Markers like `<!-- confluence-page-id: 00000000000 -->` also have case-sensitive semantics. In my opinion, it would make sense for all of these processing instruction comments to behave in a similar way.

md2conf/text.py
```diff
@@ -52,3 +56,34 @@ def wrap_text(text: str, line_length: int = 160) -> str:
...
+    <!-- md2conf-skip-start --> and <!-- md2conf-skip-end -->
+    
+    :param text: Raw Markdown text
+    :returns: Markdown text with excluded sections removed
+    """
+    # Pattern to match excluded sections
+    # (?s) is inline flag for re.DOTALL - makes . match newlines
+    # .*? is non-greedy to handle multiple exclusion blocks
+    # \s* allows optional whitespace around marker names
+    pattern = r'<!--\s*md2conf-skip-start\s*-->.*?<!--\s*md2conf-skip-end\s*-->'
```
hunyadi: I think it has to be more complex than this. Consider, for example, the proposed updates to `README.md`. They won't correctly show up in Confluence because these processing instructions will activate in the fenced code block. In contrast, consider the function `unique_title` in `toc.py`, which maintains a simple parsing state (outside/inside code block) to ensure that shell line comments like `# this is a comment` are not identified as Markdown headings when inside a fenced code block. We would need to employ a similar mechanism.

I don't know your exact use case but I would make `<!-- md2conf-skip-start -->` and `<!-- md2conf-skip-end -->` operate at the block level only. In other words, if you write these in the middle of a paragraph, they shouldn't activate. This will prevent Markdown paragraphs like

```
Content between `<!-- md2conf-skip-start -->` and `<!-- md2conf-skip-end -->` markers is removed before conversion and will not appear on the Confluence page.
```

from turning into the Confluence output

```
Content between `` markers is removed before conversion and will not appear on the Confluence page.
```

which would be confusing to readers (and likely harder to detect).

md2conf/text.py
```diff
@@ -5,6 +5,10 @@
 
 :see: https://github.com/hunyadi/md2conf
 """
+import re
+import logging
+    
+logger = logging.getLogger(__name__)
```
hunyadi: By convention, this should be:

```
LOGGER = logging.getLogger(__name__)
```

md2conf/text.py
```diff
@@ -52,3 +56,34 @@ def wrap_text(text: str, line_length: int = 160) -> str:
...
+    # Warn about unmatched markers
+    start_pattern = r'<!--\s*md2conf-skip-start\s*-->'
+    end_pattern = r'<!--\s*md2conf-skip-end\s*-->'
+    start_count = len(re.findall(start_pattern, text, flags=re.IGNORECASE))
+    end_count = len(re.findall(end_pattern, text, flags=re.IGNORECASE))
+    if start_count != end_count:
+        logger.warning(
+            f"Unmatched md2conf-skip markers: found {start_count} start marker(s) "
+            f"and {end_count} end marker(s). Content may not be excluded as expected."
+        )
```
hunyadi: I think this is an error, not a warning. Exclusion tag mismatches lead to completely garbled output, e.g. half of the document may disappear in Confluence. Markdown documents with mismatched tags should raise an exception.

tests/source/skip_nodes.md
```diff
@@ -0,0 +1,22 @@
...
+<!-- md2conf-skip-start -->
+Ignore that part
+
+Ignore this one too
+<!-- md2conf-skip-end -->
+Carry on
+
+## Section with inline
+
+As an end user, do the following.<!-- md2conf-skip-start --> As a developer you will do that instead.<!-- md2conf-skip-end --> Launch the program
```
hunyadi: This looks artificial, which is OK in a test case. However, can you share a real-world use case when you need to rely on inline exclusion? This would help assess the utility of this feature vs. the increased risk it brings.

md2conf/text.py
rpioger: Fair. I'll change that.

md2conf/text.py
rpioger: good catch. I do use it anymore

md2conf/text.py
rpioger: My use case was to not be too much restrictive in the keyword. 
But I reckon this logic shall comply with the other markers. I did not realize myself the other markers were case sensitive

md2conf/text.py
rpioger: I first thought of an error.
With the tests I ran, I figured that a mismatch just led to a full sync with Confluence. I did not want to block the user at that time.

I'll make it an error

md2conf/text.py
rpioger: I did not think about the case you pointed out.
In my mind, the writer should be aware of what he's doing should the spec be written this way. It's like getting a new car, knowing how to drive it and still deciding to go against a wall.

Now, we can enrich the pattern to consider the quotes and enforce the use of blank space around these markers. That might complexify the edge cases.

My use case was within lists, referencing some elements I did not want to be disclosed in Confluence. To respect the right format, I managed to handle inline markers. At least it was the easiest way I found.

tests/source/skip_nodes.md
rpioger: See https://github.com/hunyadi/md2conf/pull/211#discussion_r2700081181

Can you elaborate on the risks you see?

md2conf/text.py
hunyadi: To some extent, I agree with the *driving a car* analogy. On the other hand, we get the issue report if something goes wrong, even if we are not to blame, which puts the strain on us. For example, the documentation clearly states that *md2conf* supports Confluence REST API v2 only. Yet, time and again, we get bug reports that it doesn't work with REST API v1.

The implementation should be defensive to catch errors early and prevent unintentional misuse. LaTeX formulas enclosed by `$` employ similar tricks to avoid activation on phrases like *I'll pay you $20*. The reason I inquired about whether we need inline exclusion was to (possibly) make our life easier: block-level exclusion would be simpler to implement. But if your use case cannot be handled with block-level markers only, this is not a viable option.

In either case, we should make sure `README.md` renders properly both as Markdown and Confluence (without missing important detail).

rpioger: Agreed.
The way it's implemented today was (I think) the easiest way as I filter out on the text itself right before anything and without taking care of blocks yet.
As you pointed out, it comes with the cons of placement anywhere.

On the other hand, I hear you about possible upcoming bugs that could arise.
I can try to move the logic in another place. I suppose that would be in the `converter.py` for block-level exclusion.

hunyadi: Depends. `converter.py` has already passed Markdown to HTML conversion, and I am not sure if [Python-Markdown](https://python-markdown.github.io/), which is the Markdown parser *md2conf* relies on, keeps comments for downstream use. The function `unique_title` in `toc.py` might be a better bet, which implements the state machine for hopping in and out of code blocks. However, if we want to have proper (defensive) behavior for the inline use case, we would likely need code that doesn't yet exist in *md2conf* so far, there is no established pattern to copy. The challenge for the inline use case is that the start/end markers can span across (converted) HTML elements, e.g. start inside a (deeply nested) `<b>` and end in a (shallower) `<p>`. In contrast, the block use case would always have a unique parent, typically a `<div>`. We can also experiment with pre-processing, e.g. replace `<!-- md2conf-skip-start -->` with `<md2conf-skip>` and `<!-- md2conf-skip-end -->` with `</md2conf-skip>`, which will make these show up in `converter.py`, where we can discard it. Obviously, authors would need to honor element nesting, or the `lxml` reader will raise cryptic warnings. (Confluence Storage Format, which is the ultimate *md2conf* output, has to validate as XML.)

rpioger: Does your comment mean the solution currently developed in the scanner is too much brutal?

As speaking of preprocessing, my code is at the very beginning of the process. I did not dive deeply in the whole implementation of md2conf so I might not have the whole picture and constraints you see.

From my view, if the user uses `<!-- md2conf-skip-start -->` and `<!-- md2conf-skip-end -->` anywhere in markdown, I can't think of a case where he would use those markers outside a consistent block. something like this
```md
This is some text **In bold but for <!-- md2conf-skip-start --> which I start the marker skip** and then I close the marker <!-- md2conf-skip-end --> outside the bold font
``` 
If he does, I think the conversion would ultimately fail, will it? 🤔

Current implementation might be an overkill but after a quick review of the `converter.py`, I see some cons of implementing the logic at the block level:
- tree traversal
- possible performances impact. I think tree traversal costs more than string operation
- Nested elements you pointed out. That one would be tricky to handle
- multiple files changes that may introduce regressions, especially if we want to apply a html tag preprocessing before passing the node to converter.py.

I'll take a look deeper next week and see alternate options. Thanks for the insights

tests/source/skip_nodes.md
codemedic: You might want to consider the following risks for the skip feature to ensure we avoid accidental information leakage from malformed markers.

1. Unbalanced Markers

    Risk: A missing or mistyped end tag causes sensitive content to be included in the output.
    ```markdown
    <!-- confluence-skip-start -->
    Sensitive internal data...
    (Missing end marker)
    ```
    Potential Mitigation: Validate marker counts and fail-fast if they don't match.

2. Nested or Interleaved Markers

    Risk: Structural ambiguity leads to logic errors or partial leakage of private data.
    ```markdown
    <!-- confluence-skip-start -->
      Outer block
      <!-- confluence-skip-start --> Inner block <!-- confluence-skip-end -->
    <!-- confluence-skip-end -->
    ```
    Potential Mitigation: Stop execution if markers remain in the filtered output, indicating a nesting issue.

3. Out-of-Order Markers

    Risk: Incorrect sequences might inadvertently publish content intended to be hidden.
    ```markdown
    <!-- confluence-skip-end --> 
    Published content mistaken for skip 
    <!-- confluence-skip-start -->
    ```
    Potential Mitigation: Detect out-of-order markers and raise an error rather than attempting a best-effort render.

Prioritizing these safety checks ensures data integrity is maintained even when users make mistakes in their Markdown structure.

md2conf/text.py
hunyadi: > As speaking of preprocessing, my code is at the very beginning of the process. I did not dive deeply in the whole implementation of md2conf so I might not have the whole picture and constraints you see.

The main issue is losing context. Even though *md2conf* doesn't currently report line/column numbers when an error occurs, this feature has been requested by several users, and our long-term plan is that we will eventually implement this. Replacing raw text with regular expressions interferes with line/column numbers in a way that is difficult to track.

> If he does, I think the conversion would ultimately fail, will it? 🤔

Yes. The challenge is that sometimes you get crazy XML parsing errors, which are difficult to decipher. For example, some of the examples I tried (and shared in issue #210) lead to not particularly helpful error messages.

> Current implementation might be an overkill but after a quick review of the converter.py, I see some cons of implementing the logic at the block level:

Performance is less of an issue. The main bottleneck is API round-trip. Sometimes the same information is retrieved multiple times with different API calls. We have some room to optimize. The slowest part of the pipeline is converting diagrams to PNG/SVG but this is handled with external software that we have no control over.

Nested elements are in my opinion easier to tackle with tree traversal than with regular expressions. Regular expressions don't maintain parsing state in a way we can interact with but an HTML tree perfectly reflects element hierarchy.

As for block-level vs. inline, my initial thought was to restrict the scope. However, based on my experiments shared in issue #210, inline is mostly OK. The limitations I found were related to lists when some list items need to be excluded but not others. (I would assume a similar problem arises with excluding parts of a block quote.) We should at least carefully document what is allowed and what is not.

rpioger: > The main issue is losing context. Even though md2conf doesn't currently report line/column numbers when an error occurs, this feature has been requested by several users, and our long-term plan is that we will eventually implement this. Replacing raw text with regular expressions interferes with line/column numbers in a way that is difficult to track.

Ok, I get the idea. Thanks for the insights. I understand why my code uses a bazooka to shoot an ant.

> Nested elements are in my opinion easier to tackle with tree traversal than with regular expressions. Regular expressions don't maintain parsing state in a way we can interact with but an HTML tree perfectly reflects element hierarchy.

I'll dive deeper in the logic to comply with the existing implementation.

tests/source/skip_nodes.md
rpioger: Hi @codemedic ,

I considered 1. and 2. with the first implementation I proposed with an error in case of counters discrepancies. Yet, at the time I wrote the code I have in mind a `Keep It Simple` with the logic of _The user shall know what he's doing_. If there is a bad marker, then md2conf still publishes and there is no harm (or should not). Should very sensitive information be written in the same Markdown, I suppose that would be a bad design if I may call that this way. Generally speaking, the user shall not mix normal doc and very sensitive information in the same document - or at least this is not how I initially thought of that feature.
3. was not properly captured though. That one can be an issue as not managed effectively.

That being said, I'm OK to enrich the cases for a more robust management of the markers. I just think that better is the enemy of good.

With both comments from you guys, I'll hold the PR and will refactor the solution

rpioger:
I should be good. The `check.bat` passed

md2conf/converter.py
```diff
@@ -1417,6 +1507,12 @@ def transform(self, child: ElementType) -> ElementType | None:
                 elif "admonition" in classes:
                     return self._transform_admonition(child)
 
+                # <div class="confluence-skip">...</div>
+                # Content marked for exclusion from Confluence (block-level). Clear the element.
+                elif "confluence-skip" in classes:
+                    child.clear()
```
hunyadi, line 1517: This will remove element descendants but not the element itself (i.e. `<div class="confluence-skip">` would remain in the tree as `<div>`). Assume we also remove the element here. Would we still need `_cleanup_empty_elements`?

md2conf/converter.py
```diff
@@ -1501,6 +1597,15 @@ def transform(self, child: ElementType) -> ElementType | None:
                 if "arithmatex" in classes:
                     return self._transform_inline_math(child)
 
+                # <span class="confluence-skip">...</span>
+                # Content marked for exclusion from Confluence (inline)
+                elif "confluence-skip" in classes:
+                    # Explicitly preserve tail before clearing.
+                    tail = child.tail
+                    child.clear()
```
hunyadi, line 1609: Same reasoning as for `<div class="confluence-skip">`. I would rather remove elements in a single pass than in multiple passes.

md2conf/converter.py
```diff
@@ -275,6 +303,68 @@ def is_placeholder_for(node: ElementType, name: str) -> bool:
...
+    :returns: HTML string with comments replaced by custom elements
+    """
+    # Pattern to match skip section markers with surrounding context
+    # Captures newlines/whitespace before and after to determine if block-level
+    start_pattern = r"<!--\s*confluence-skip-start\s*-->"
+    end_pattern = r"<!--\s*confluence-skip-end\s*-->"
+
+    # Count markers for validation
+    start_count = len(re.findall(start_pattern, html))
+    end_count = len(re.findall(end_pattern, html))
```
hunyadi, line 302: Naively, I would think these do not match if we have a single `<!-- confluence-skip-end -->` in a fenced code block, which we should ignore.

tests/source/skip_nodes.md
```diff
@@ -0,0 +1,37 @@
...
+- Server Installation
+<!-- confluence-skip-end -->
+- FAQ
+
+```python
+def foo():
+    print("Using skip markers in code block is")
+    <!-- confluence-skip-start -->
+    print("ignored")
+    <!-- confluence-skip-end -->
```
hunyadi: I would recommend we remove `<!-- confluence-skip-start -->` or `<!-- confluence-skip-end -->` to show that even unpaired tags don't cause any trouble. Plus, I would put a Python comment in front to ensure the code validates as Python:

```python
# <!-- confluence-skip-end -->
```

md2conf/converter.py
hunyadi: Looking at test runs, it appears that Python-Markdown doesn't emit CDATA sections for code blocks but uses HTML escapes instead. As such, the HTML comments in the code block become:

```
&lt;!-- confluence-skip-end --&gt;
```

This means that `transform_skip_comments_in_html` won't count these as occurrences.

md2conf/converter.py
rpioger: I tried some things to avoid parsing the tree multiple times and ended up with _out of range_ issues and discrepancies.
I completely align that passing only once is better but I prefer not to touch the visitor implementation.

md2conf/converter.py
rpioger: See https://github.com/hunyadi/md2conf/pull/211#discussion_r2713447885

md2conf/converter.py
hunyadi: OK, that's totally fine. I am going to add some minor tweaks (e.g. short-circuit text manipulation if there is no `<!-- confluence-skip-start -->`, or raise an error when there is a tag count mismatch, and then merge the pull request. We can simplify the implementation as a follow-up work item.

Issue #213, opened 2026-01-17
[Feature] front-matter support to mimic Confluence Cloud scheduled updates


codemedic:
# Feature Request: front-matter support to mimic Confluence Cloud scheduled updates

## Describe your idea

Confluence Cloud supports *scheduled page updates* in the Web UI. Conceptually, this allows you to keep editing a page and have the *latest version* become visible once a certain condition is met, without requiring an exact execution time.

There’s no public REST API for this feature, so when publishing from GitHub we currently have to approximate the behaviour externally. While Confluence Cloud allows specifying an exact time, we can leverage GitHub Actions (e.g., cron schedules or polling) to achieve a similar effect.

### Proposed Idea

Allow an optional front-matter field, for example:

```yaml
publish_after: 2026-01-18T09:00:00Z
```

- If the condition is not met → skip publishing
- If the condition is met → publish the current content

This isn’t about exact timing, but about modelling the same “pending update → latest content goes live” semantics that Confluence Cloud provides in the UI. It is up to the workflow model to decide whether to publish from tagged / approved changes etc.

### Why this helps

- Makes GitHub-driven publishing behave more like the Confluence Cloud editor.
- Works well with polling-style workflows (e.g., Actions running every N minutes).
- Avoids relying on precise cron execution for every single page update.
- Keeps the logic close to the content, rather than the CI configuration.
- Supports "publish on X date at start of working hours" (e.g., Action runs daily at 9:00 AM, with `publish_after` set to 8:59 AM).

### CLI options

- A CLI flag to bypass or ignore the `publish_after` setting (e.g., `--ignore-publish-after`), allowing immediate synchronization of all pages regardless of their scheduled date. This helps with previewing changes from a branch, for instance.

codemedic:
Hey @hunyadi,

I really like this direction. Even though it feels like a bit of an architectural shift for a relatively specific feature, I definitely see the long-term value in providing a hook point to make `md2conf` extensible. It opens up a lot of possibilities.

How do you envision the 'pluggable' part working in practice? I can see a few paths:

* Python-native: An extension configured and loaded via `pyproject.toml`. This assumes the user is comfortable with Python semantics, which feels like a fair trade-off for this tool.
* CLI-based: A command-line argument to load a local plugin file.
* Language-agnostic: An external binary hook, which would remove the Python requirement for the plugin itself.

[...]

PR #214, opened 2026-01-17
feat: Add scheduled publishing functionality

codemedic:
Implements #213

### DIFF
--- a/tests/test_publish_after.py
+++ b/tests/test_publish_after.py
@@ -1,3 +1,10 @@
+"""
+Tests for a use-case of the Selective Synchronization feature.
+
+This test demonstrates how to implement scheduled document updates using a custom predicate,
+as described in the README.
+"""
+
 import unittest
 from datetime import datetime, timezone
 from pathlib import Path
--- a/tests/test_sync_by_env.py
+++ b/tests/test_sync_by_env.py
@@ -1,3 +1,10 @@
+"""
+Tests for a use-case of the Selective Synchronization feature.
+
+This test demonstrates how to synchronize documents based on target environment,
+as described in the README.
+"""
+
 import unittest
 from pathlib import Path
 from tempfile import TemporaryDirectory


tests/test_publish_after.py
```diff
@@ -0,0 +1,127 @@
...
+            self.assertTrue(node.synchronized)
+
+    def test_publish_after_ignore_bypass(self) -> None:
+        with TemporaryDirectory() as tmp_dir:
+            tmp_path = Path(tmp_dir)
+            md_file = tmp_path / "test.md"
+            with open(md_file, "w") as f:
+                f.write("---\npublish_after: 2099-01-01T00:00:00Z\n---\nBody")
+
+            options = DocumentOptions(synchronize_if=publish_after, params={"ignore_publish_after": "True"})
```
hunyadi, line 108: I think you are making this more complicated than it should be. `params` should be part of the predicate you are passing. [Partial functions](https://docs.python.org/3/library/functools.html#functools.partial) may be useful to accomplish this but a plain lambda or Python function with a variable capture would do.

README.md
```diff
@@ -566,6 +566,135 @@ This Markdown document is neither parsed, nor synchronized with Confluence.
 
 This is useful if you have a page in a hierarchy that participates in parent-child relationships but whose content is edited directly in Confluence. Specifically, these documents can be referenced with relative links from other Markdown documents in the file system tree.
 
+### Selective Synchronization
```
hunyadi, line 569: I think we might want to document this feature elsewhere, e.g. in a separate `docs` folder. Today, `README.md` is about 800 lines. This change-set would add almost 150 lines. This means that about 15% of the core documentation would talk about what I consider a relatively niche feature.

md2conf/__main__.py
```diff
@@ -59,6 +60,8 @@ class Arguments(argparse.Namespace):
     alignment: Literal["center", "left", "right"]
     max_image_width: int | None
     use_panel: bool
+    synchronize_if: str | None
+    params: dict[str, Any]
```
hunyadi, line 64: I don't think we need this as a command-line argument to core *md2conf*. This is information your utility or extension would acquire from the user (e.g. with your own command-line parser, the environment, or simply a variable defined in a Python file), attach to the predicate (a.k.a. partial function), and then pass only this predicate.

md2conf/__main__.py
```diff
@@ -59,6 +60,8 @@ class Arguments(argparse.Namespace):
     alignment: Literal["center", "left", "right"]
     max_image_width: int | None
     use_panel: bool
+    synchronize_if: str | None
```
hunyadi, line 63: I don't think we should be reading predicates as strings from the command line, which can open up the tool to security vulnerabilities. Rather, developers should define this predicate in their own Python script, in which they also instantiate a `Publisher` object, and then pass the predicate to `Publisher`.

md2conf/__main__.py
```diff
@@ -326,9 +329,51 @@ def get_parser() -> argparse.ArgumentParser:
...
+    :returns: The imported callable.
+    """
+
+    try:
+        module_name, callable_name = name.split(":")
+    except ValueError:
+        raise ArgumentError(f"invalid synchronize-if format: {name}. Expected 'module:callable'") from None
+
+    try:
+        module = importlib.import_module(module_name)
```
hunyadi, line 362: This raises security concerns similar to `eval()`. The [documentation for eval](https://docs.python.org/3/library/functions.html#eval) writes:

> Warning This function executes arbitrary code. Calling it with user-supplied input may lead to security vulnerabilities.

The same principle applies to this approach.

hunyadi (changes requested):

md2conf/__main__.py
codemedic: I appreciate the rigorous look at the security implications. You are correct that `importlib` and `eval()` share a similar risk profile regarding arbitrary code execution. 

However, in the context of a CLI tool, I believe we should differentiate between **vulnerability** (an unintended oversight) and **extensibility** (a deliberate design choice for modularity).

1. **The Trust Boundary:** For a CLI utility, the security boundary is effectively the local filesystem. To exploit the `importlib` call, an attacker would first need the privilege to write a Python module to the host disk or the CI runner. If an attacker achieves that level of access, the environment is already fully compromised; the ability to hook into `md2conf` adds no marginal utility to the exploit.

2. **Interface Stability & Coupling:** Requiring users to move to the Python library interface (as detailed in the README) significantly increases the coupling between user logic and the internal API. A downstream script becomes a 'maintenance hostage' to any refactor of `ConnectionProperties`, `DocumentOptions`, or the `Publisher` class. 

3. **Operational Security vs. Complexity:** One could argue for 'Safe-Load' checks (e.g., verifying file ownership or permissions before importing). However, I would dismiss this as overkill for `md2conf`. It adds unsubstantiated complexity to the core for a threat model that essentially assumes the user's OS-level security has already failed.

I understand that a feature like `publish_after` might be considered niche within the broader scope of the project. However, maintaining this extension point at the CLI level provides a stable, high-level abstraction that preserves the tool's flexibility for the community's power users without cluttering the core logic. 

If the primary concern is the optics of the security risk, perhaps a balanced path forward would be to keep the CLI entry point but make this 'opt-in' nature explicit via a `SecurityWarning` in the help text? I’d be happy to add that warning to ensure we’re being proactive about security best practices, while still offering the flexibility that makes `md2conf` such a capable tool for different workflows.

md2conf/__main__.py
codemedic: Please see the comment below. https://github.com/hunyadi/md2conf/pull/214#discussion_r2714849172

md2conf/__main__.py
codemedic: I agree that for users interacting directly with the Python API, `partial` functions or closures are the idiomatic way to handle state. 

However, looking at the CLI as the primary interface for automation, the `params` map serves as a configuration bridge. Since a CLI user cannot 'partially apply' a Python function from a shell command, they require a way to inject data into their predicates.

Apart from implementing a more complex external configuration file, the only alternatives for a CLI-driven workflow would be:
* **Hard-coding environment-specific values** directly into the Python predicate module, which limits reusability.
* **Relying on `os.environ`** inside the predicate to fetch external state, which moves configuration into the global scope and makes it harder to debug/trace.

The `params` map provides a structured, explicit 'data transport' layer that keeps the CLI as the source of truth. Does this distinction between the library's functional approach and the CLI's need for a configuration bridge help clarify the intent?

README.md
codemedic: I couldn't agree more. A 15% increase for a specialized feature is a lot of weight for a single `README.md` to carry. 

I actually looked into rehashing the documentation to make the landing page more approachable for newcomers, but I was concerned about potentially breaking the PyPI landing page. 

Since the existing `/docs` folder appears to be git-ignored for auto-generated content, I’m happy to move the detailed examples and the `synchronize_if` technical reference into a new dedicated directory (perhaps `/manual/` or `/examples/`) or just a standalone file like `ADVANCED_USAGE.md` in the root.

We can then keep the `README.md` lean by adding a small 'Advanced Features' section that links out to the new page. This preserves the main README as a clean 'Quick Start' while giving power users a dedicated space for complex logic.

Would you prefer a specific naming convention for a new documentation folder, or should I just start with a standalone file in the root for now?

tests/test_publish_after.py
codemedic: I think I have covered this in the comment below.
https://github.com/hunyadi/md2conf/pull/214#discussion_r2714892266

README.md
hunyadi: Links don't work in PyPI because it only displays `README.md` and nothing else. However, putting a short summary of how you can assign a predicate to `synchronize_if` in the options you pass to `Publisher` would be sufficient. Detailed explanation could go into a separate folder like `manual`. Please use lowercase for file names, e.g. `advanced_usage.md`. `docs` is in `.gitignore` at the moment because I have a documentation generator tool that produces output by extracting doc-strings from Python modules and classes but I don't publish them anywhere. GitHub has a great website view but it requires writing a special GitHub Actions script. I have the script integration for other projects that I own but not yet for *md2conf*.

md2conf/__main__.py
hunyadi: I understand the intent but I don't like the implementation approach, I am afraid. I would delegate this task (and responsibility) to developers writing tools. This decoupling would allow developers who want to expose specialized features to move forward at their own pace, and keep *md2conf* focused on core tasks.

Specifically, let's suppose you are writing a utility that supports synchronizing Markdown files based on an activation date. This utility can have its own command line. The parser may import the command-line parser from *md2conf* and extend it with new arguments. One command line argument might be of type `datetime`, which takes an ISO date-time string. The value of the argument would be captured by the predicate that is assigned to `synchronize_if` in the options you pass to `Publisher`.

The important point is that only `synchronize_if` lives in the *md2conf* repository. Everything else lives in a dedicated repository, relying on extension and inheritance to copy behavior from *md2conf*. If you decide that you want to capture more complex conditions beyond date, you can accomplish this entirely on your own, you don't have to come to an agreement with *md2conf* maintainers.

md2conf/__main__.py
hunyadi: Interface stability and coupling are the primary reasons. We are taking steps to make the API external. Given the external-facing *md2conf* API, developers are able to write their own extensions and tools, which may have a command-line interface. This is commonly referred to as *API-first* design: library developers write reusable APIs and application developers would write tooling that combines API features in novel ways. We are going to continue including a command-line interface in *md2conf* but it is going to support only a subset of the features available via the API.

codemedic:
I’ve given this some thought, and I think our visions for the project's architecture are headed in different directions. I've decided to close this PR for now.

I hear your ambitions for `md2conf` as a library, and I truly wish you the best of luck with that roadmap. From my perspective, however, the project’s greatest potential lies in being a well-balanced, CLI-first tool for both local use and CI/CD environments. I see it as a platform that could benefit from a dynamic extension interface—much like how HTTPie or flake8 allow users to provide custom logic via installed plugins without requiring them to maintain a custom Python wrapper for the entire library.

For users who prefer the stability and simplicity of a CLI interface over maintaining custom Python integration code, that 'plug-and-play' extensibility is the missing piece. Since I’m keen to see the tool move in that direction and I've already invested significant time here, I’m going to continue developing these features in my own fork for the time being.

I’ve enjoyed the technical exchange and look forward to potentially collaborating on other features in the future. Over and out for now!

Issue #205, opened 2026-01-13
[Feature] Allow override api url


shkup:
### Describe your idea

Our confluence in the data center doesn't support REST V2. please allow to explicitly override it so it wall call  `rest/api/content` rather than `api/v2/pages`.  Right now `--api-url` changes the base `url`, but `md2conf` still appends `api/v2/pages` internally.

Issue #168, opened 2025-10-03
md2conf uses Cloud API v2 for on-prem Confluence even without --api-url


kpallade:
**Describe the bug**
md2conf (v0.4.7) calls Cloud API v2 endpoints (`api/v2`) when used with an on-premise Confluence, even if `--api-url` is not provided. This causes 401 errors since on-prem only supports the REST API v1 (`/rest/api/...`).

**To Reproduce**
Run md2conf on an on-premise server:

```
md2conf "index.md"
--domain "confluence.company.fr"
--path "/"
--username "user"
--api-key "personal_access_token"
--space "ABC"
```

**Expected behavior**
md2conf should detect on-premise and use the v1 API (`/rest/api/...`) instead of Cloud v2.

Issue #141, opened 2025-07-21
How do I trigger generation of Confluence macros?


grepwood:
Hello! I'm really happy with md2conf and I've been using it a lot recently. It's fantastic to work with.

One of my use cases now involves the embedding of entire HTML pages into the Markdown file.

Here's what I can generate thus far, before I feed it to `md2conf`:

````
```html
<!doctype html>
<html class="h-full bg-gray-100" lang="en">
<head>
</head>
<body>
</body>
</html>
```
````

What I would like `md2conf` to generate, when it creates a `.csf` file is this:
```
<ac:structured-macro ac:name="html" ac:schema-version="1"><ac:plain-text-body><![CDATA[<!doctype html>
<html class="h-full bg-gray-100" lang="en">
<head>
</head>
<body>
</body>
</html>
]]></ac:plain-text-body></ac:structured-macro>
```

Is there any way to do this with `md2conf` 0.2.7? The way I wrote my Markdown, it'll only embed the HTML as code.

Issue #110, opened 2025-05-21
Does md2conf work with Data Center instead of Cloud?


Jerakin:
I tried this with our confluence server and always get errors. Is it supposed to work with servers or are only cloud supported?

```
2025-05-21 14:42:28,751 - ERROR - main [242] - 401 Client Error:  for url: https://my-url/api/v2/pages/2998052547?body-format=storage
```

hunyadi:
You might want to switch to *md2conf* version [0.2.7](https://pypi.org/project/markdown-to-confluence/0.2.7/).

Unfortunately, Confluence REST API comes in two (not interchangeable or backwards-compatible) versions: v1 and v2, and Data Center supports only v1, and Cloud supports only v2. More on this in issue #93. Our main use case is Confluence Cloud, and we don't have a Data Center deployment to test *md2conf* against. However, users have reported that md2conf 0.2.7 works fine with Data Center.

Some users may need to pass `--webui-links` to make sure appropriate cross-page links are generated.

Jerakin:
Thanks for the input. I have been toying with the idea of forking the repository and re-implementing the data center api, but that's not something I have time to do currently. Let me know if you (or someone else reading this) creates one.

Another option I was thinking of was to create two separate interfaces (one for cloud and one for server) and then pass correct interface to the `ConfluenceSession` which would work the abstractions instead. However, it seems like supporting server interface is out of scope for this library.

hunyadi:
Internally, we use Confluence Cloud, and – unfortunately – we don't have a Data Center deployment. Without the ability to run the integration test suite against the appropriate Confluence environment, we can't assure quality, which is why we decided to remove those v1 endpoint calls from `api.py` that are no longer supported in Confluence Cloud. If someone were to re-add v1 (and therefore Data Center) support, they would have to look for `VERSION_2` in `ConfluenceSession`, and provide a Data Center compatible implementation for these calls (that ultimately make a `VERSION_1` API call). The v1 and v2 data model (e.g. pages, spaces, etc.) are mostly compatible with only a few caveats. Some functionality in Confluence remains accessible only via the v1 API, which is why *md2conf* has a mixture of v1 and v2 API calls.

The last version of *md2conf* that had v1 API calls only is version [0.2.7](https://pypi.org/project/markdown-to-confluence/0.2.7/).

PR #222, opened 2026-02-03
feat: add REST API v1 support - Confluence Data Center

ojacques:
This PR adds support for Confluence REST API v1, in other words Confluence Data Center. This allows to use all the latest features of md2conf, also for Confluence Data Center just by adding the option `--api-version v1`

   Addresses 
   #82
   #93
   #110
   #141
   #168
   #205
   
   _*Note on the approach*_ : I am conscious this PR adds a lot more code (actually more unit tests than actual code, but still). I paid attention to not change v2 code paths as much as possible. I extensively tested this with both on-premises and [Confluence server Docker image](https://hub.docker.com/r/atlassian/confluence) and a 180 pages markdown/mkdocs site with code, attachments, admonitions. Also added unit tests as this is an important consideration for @hunyadi - given that you don't use Confluence Data Center. Static tests are passing too.
   
   Hopefully, you are willing to consider merging it in 😉

...and thanks for this great tool!


ojacques:
Thanks for the swift reply, as always.

> From an implementation perspective, rather than jumping in and out of REST API v1 blocks, I would split v2 and v1 implementations into separate classes. 

That makes sense. Let me give it a try. 

> Please separate out any changes unrelated to API v1 support such as extra admonitions. These unrelated changes can be merged separately in a much more straightforward process. In contrast, API v1 support might be a longer journey, and keeping the change-set as small as possible can speed it up.

Done. See #223 .

md2conf/__main__.py
```diff
@@ -39,6 +39,7 @@ class Arguments(argparse.Namespace):
     username: str | None
     api_key: str | None
     space: str | None
+    api_version: str
```
hunyadi: You could define this as `Literal["v1", "v2"]` in which case you would not need to check for invalid values.

README.md
```diff
@@ -660,6 +660,8 @@ options:
   -a, --api-key API_KEY
                         Confluence API key. Refer to documentation how to obtain one.
   -s, --space SPACE     Confluence space key for pages to be published. If omitted, will default to user space.
+  --api-version {v1,v2}
+                        Confluence REST API version to use (v1 for Data Center/Server, v2 for Cloud). Default: v2
```
hunyadi: I am an enthusiastic advocate of consistency. This should be `(default: v2)` to align with how default values are declared elsewhere.

README.md
```diff
@@ -727,6 +729,7 @@ properties = ConnectionProperties(
     user_name=str() or None,
     api_key=str(),
     headers={str(): str()} or None,
+    api_version='v1' or 'v2',  # Default: 'v2'
```
hunyadi: For the rest of the options, the default value comes first.

README.md
```diff
@@ -766,9 +769,19 @@ with ConfluenceAPI(properties) as api:
 
 ### Confluence REST API v1 vs. v2
 
-*md2conf* version 0.3.0 has switched to using [Confluence REST API v2](https://developer.atlassian.com/cloud/confluence/rest/v2/) for API calls such as retrieving current page content. Earlier versions used [Confluence REST API v1](https://developer.atlassian.com/cloud/confluence/rest/v1/) exclusively. Unfortunately, Atlassian has decommissioned Confluence REST API v1 for several endpoints in Confluence Cloud as of due date March 31, 2025, and we don't have access to an environment where we could test retired v1 endpoints.
+*md2conf* supports both [Confluence REST API v1](https://developer.atlassian.com/cloud/confluence/rest/v1/) and [Confluence REST API v2](https://developer.atlassian.com/cloud/confluence/rest/v2/). By default, *md2conf* uses API v2, which is the go to version for Confluence Cloud. By setting `--api-version v1`, you can use *md2conf* with Confluence Data Center / On-premises.
```
hunyadi: I would be cautious about the exact wording. We can't officially support REST API v1 until we have a working operational environment. I tried downloading and running the Docker image below but it asks for a license code when it starts up:

```
docker run -d -p 8090:8090 atlassian/confluence
```

I would say we continue to support REST API v2, we no longer mention version 0.2.7 and we offer REST API v1 integration as *best effort* and/or *community supported*.

tests/test_v1_v2_space_id_validation.py
hunyadi, line 1: I am not sure if we need this test file any longer. Both v2 and v1 implementations now have the same method signatures, which mostly expect a `space_id`, even though it is identical to a `space_key` behind the scenes for v1. A previous version of `api.py` relaxed `space_id` to be optional but I restored `space_id` as a mandatory parameter such that we don't need to update code outside of `api.py`.

md2conf/api.py
```diff
@@ -393,7 +395,7 @@ def __enter__(self) -> "ConfluenceSession":
         if self.properties.headers:
             session.headers.update(self.properties.headers)
 
-        self.session = ConfluenceSession(
+        self.session = ConfluenceSessionV2(
```
hunyadi: This is where we will need to instantiate either `ConfluenceSessionV2` or `ConfluenceSessionV1`, depending on the value of `api_version`.

md2conf/publisher.py
```diff
@@ -177,7 +177,13 @@ def _synchronize_subtree(self, node: DocumentNode, parent_id: ConfluencePageID,
 
             update = True
 
-        space_key = self.api.space_id_to_key(page.spaceId)
+        # For v1, page.spaceId already contains the space key (not ID)
+        # For v2, page.spaceId contains the space ID and needs conversion
+        if self.api.api_version == ConfluenceVersion.VERSION_1:
+            space_key = page.spaceId  # Already a space key for v1
+        else:
+            space_key = self.api.space_id_to_key(page.spaceId)
```
hunyadi: These changes are no longer needed. `space_id_to_key(...)` is a no-op for v1 but this is opaque to the interface user.

md2conf/publisher.py
```diff
@@ -329,7 +335,12 @@ def _get_unique_title(self, document: ConfluenceDocument, path: Path) -> str | N
         meta = self.page_metadata.get(path)
         if meta is not None and meta.title != title:
             # title has changed, check if new title is available
-            page_id = self.api.page_exists(title, space_id=self.api.space_key_to_id(meta.space_key))
+            # meta.space_key contains the actual space key for both v1 and v2
+            # Pass it as space_id parameter (which accepts key for v1, ID for v2)
+            if self.api.api_version == ConfluenceVersion.VERSION_1:
+                page_id = self.api.page_exists(title, space_id=meta.space_key)
+            else:
+                page_id = self.api.page_exists(title, space_id=self.api.space_key_to_id(meta.space_key))
```
hunyadi: These changes are no longer needed. `space_key_to_id(...)` is a no-op for v1 but this is opaque to the interface user.

tests/test_v1_api_methods.py
```diff
@@ -0,0 +1,218 @@
...
+from unittest.mock import MagicMock, patch
+
+from md2conf.api import (
+    ConfluenceError,
+    ConfluenceSession,
+    ConfluenceVersion,
+)
+
+
+class TestV1PageMethods(unittest.TestCase):
```
hunyadi: I don't think we need this test case any longer as routing via inheritance is part of the Python language.

tests/test_v1_v2_api_version.py
```diff
@@ -0,0 +1,142 @@
...
+"""
+
+import unittest
+from unittest.mock import MagicMock, Mock, patch
+
+from md2conf.api import ConfluenceVersion
+from md2conf.environment import ArgumentError, ConnectionProperties
+
+
+class TestAPIVersionParsing(unittest.TestCase):
```
hunyadi: I believe that if we define `api_version` as `Literal["v1", "v2"]` none of these tests in this case should be necessary.

tests/test_v1_v2_api_version.py
```diff
@@ -0,0 +1,142 @@
...
+class TestURLConstruction(unittest.TestCase):
+    """Test URL construction for different API versions."""
+
+    def test_v1_url_prefix(self) -> None:
+        """Test that v1 URLs use /rest/api prefix."""
+        self.assertEqual(ConfluenceVersion.VERSION_1.value, "rest/api")
+
+    def test_v2_url_prefix(self) -> None:
+        """Test that v2 URLs use /api/v2 prefix."""
+        self.assertEqual(ConfluenceVersion.VERSION_2.value, "api/v2")
```
hunyadi: This seems noise. The value of an enumeration constant is trivially its value, which is how `enum` works in Python. The purpose of test cases is to catch non-trivial errors.

hunyadi (changes requested):
Thank you for contributing Confluence REST API v1 support to *md2conf*!

I have made a few comments to help align with the existing coding style, and reduce the volume of code future maintainers need to support.

Once these minor issues are addressed, the pull request is ready to merge into the mainline branch.

Please verify if everything works as expected if you plug in `ConfluenceSessionV1` in place of `ConfluenceSessionV2`. (The latest version of the code accesses functionality via the abstract base class `ConfluenceSession`.)

ojacques:
Thanks! With your latest changes, I found that api_version was not actually used. See if how I fixed it suits you. Also had a go at the `Content-Type` headers, and it works, but I am not sure this the way it should be. It could be out of scope of support API v1 though.
Other than that, tested with a 252 pages site, and the basics are there. I have issues with the parser which does not tolerate incorrectly formed html (happens when writing docs), but I can / should workaround that by running a script to fix issues - before publication.

hunyadi:
> Thanks! With your latest changes, I found that api_version was not actually used. See if how I fixed it suits you.

If you make `api_version` a `Literal["v1", "v2"]`, you can turn the conditional into a `match` that type checkers can verify for exhaustiveness. (This doesn't apply to a series of `if`/`elif`/`else` statements.) However, these are minor technical issues that I can address. The important point is to make sure that the *v1* implementation (as in this pull request) runs without error on your system, which is something I cannot verify.

> Also had a go at the `Content-Type` headers, and it works, but I am not sure this the way it should be. It could be out of scope of support API v1 though.

Technically, what you wrote is not strictly valid. `Content-Type` determines what type of payload the HTTP GET request has. However, an HTTP GET request should never have a payload (and doesn't have as per the implementation in `_get`), so `Content-Type` doesn't make much sense. (We are setting a property on something that doesn't exist.) On the other hand, Confluence Data Center may be buggy, and may require this HTTP header. I will change it to apply to the *v1* implementation only, since I don't want to introduce conceptually wrong behavior for the officially supported version. (Many HTTP client implementations will remove `Content-Type` if you set it on an HTTP GET request for the reasons above.)

> Other than that, tested with a 252 pages site, and the basics are there.

That's awesome. I had made more extensive changes w.r.t. how page details and attachments are handled but if they work fine, we are good to go.

> I have issues with the parser which does not tolerate incorrectly formed html (happens when writing docs), but I can / should workaround that by running a script to fix issues - before publication.

Unfortunately, the Markdown/HTML stack we currently use doesn't offer much help in this regard, your final document must be well-formed XHTML. I have experimented with replacing Python-Markdown with a SAX-style parser called [md4c](https://github.com/mity/md4c), which would allow better error messages (and perhaps recovery) but this would entail more substantial work, which we don't have the capacity for at the moment.

I will make some further changes to ensure that typing/linting checks and unit test succeed (see `check.sh` or `check.bat`, they currently fail) and will merge the pull request on February 15.

hunyadi (approved):

PR #223, opened 2026-02-04
fix: more admonitions - mkdocs material

ojacques:
Add more admonitions mapping, to support the one from [mkdocs material](https://squidfunk.github.io/mkdocs-material/reference/admonitions/).

### DIFF
--- a/md2conf/converter.py
+++ b/md2conf/converter.py
@@ -860,13 +860,20 @@ class ConfluenceStorageFormatConverter(NodeVisitor):
         else:
             admonition_to_csf = {
                 "attention": "note",
+                "abstract": "note",
+                "bug": "warning",
                 "caution": "warning",
                 "danger": "warning",
                 "error": "warning",
+                "example": "info",
+                "failure": "warning",
                 "hint": "tip",
                 "important": "note",
                 "info": "info",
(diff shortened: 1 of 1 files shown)


hunyadi (changes requested):
Thanks for adding these new admonition types!

Please also check `ConfluencePanel.from_class`. When the configuration option `use_panel` is enabled, we activate the function `_transform_panel`, which emits a `structured-macro` of type `panel` with custom icon and background color. I believe some of the *mkdocs* admonition types like *bug* or *quote* could have an informative icon of their own. Even if you choose to duplicate one of the existing definitions, you must add the new admonition types to `ConfluencePanel.from_class` such that we don't get an exception when `use_panel` is enabled.

I think you might have accidentally forgotten to move `tests/test_admonition_mappings.py` from your earlier pull request. I don't insist on migrating that test as I believe the existing test infrastructure already covers these cases, so long as the existing tests are updated with the new values. Specifically, you need to update `tests/source/admonition.md`, `tests/target/admonition.xml` (to cover the variant when `use_panel` is disabled) and `tests/target/panel.xml` (when `use_panel` is enabled) such that we have full test coverage of these new types.

Issue #226, opened 2026-02-09
[Bug] Some languages are not supported in code blocks


nikaltipar:
### Latest version check

- [x] I confirm I can reproduce the issue with the **latest version** of *md2conf*.

### Describe the issue

I am trying to add some documentation for my [terraform project](https://github.com/NVIDIA/slack-to-jira) that uses a lot of HCL blocks and thus I need to propagate some of those to my internal documentation. I've seen that there is a map for language code blocks in [md2conf](https://github.com/hunyadi/md2conf), but the list is non-exhaustive. On this occasion, HCL is not supported.

What I propose:

- Change the `LANGUAGES` map to have only handling for non-default options.
  In the current form it would be something like
   ```
    _LANGUAGES = {
        "csharp": "c#",
        "javascript": "js",
        "python": "py",
   }
   ``` 
- Use "passthrough" for all values not on the map, e.g. `_LANGUAGES.get(language_name, language_name)`

PR #227, opened 2026-02-10
Option: allow unsupported languages in code blocks

nikaltipar:
* Add a new parameter to allow unsupported languages in provided code blocks.
* Originally (and now in the default case), only languages drawn from a given map are allowed in code blocks. Any other option is ignored and "none" is chosen as default.
* With -no-force-valid-language, any language provided in code blocks but not belonging as key in the language map, will be passed as is to confluence.

#226

### DIFF
--- a/md2conf/converter.py
+++ b/md2conf/converter.py
@@ -808,9 +808,7 @@ class ConfluenceStorageFormatConverter(NodeVisitor):
 
             language_id = _LANGUAGES.get(
                 language_name,
-                None
-                if self.options.force_valid_language
-                else language_name,
+                None if self.options.force_valid_language else language_name,
             )
         else:
             language_id = None
--- a/tests/test_conversion.py
+++ b/tests/test_conversion.py
@@ -519,25 +519,6 @@ class TestConversion(TypedTestCase):
 
         self.assertEqual(actual, expected)
 
-    def test_unknown_code_language_default(self) -> None:
-        """Unknown code block languages are emitted as language 'none' when force_valid_language is True (default)."""
-        converter_options = ConverterOptions(
-            force_valid_language=True,
-        )
-        _, doc = ConfluenceDocument.create(
-            self.source_dir / "unknown_code_language.md",
-            ProcessorOptions(converter=converter_options, line_numbers=False),
-            self.source_dir,
-            self.site_metadata,
-            self.page_metadata,
-        )
-        actual = standardize(doc.xhtml())
-
-        with open(self.target_dir / "unknown_code_language.xml", "r", encoding="utf-8") as f:
-            expected = substitute(self.target_dir, f.read())
-
-        self.assertEqual(actual, expected)
-
     def test_unknown_code_language_passthrough(self) -> None:
         """With force_valid_language=False, unknown language names are passed through."""
         converter_options = ConverterOptions(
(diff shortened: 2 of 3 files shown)


tests/test_conversion.py
```diff
@@ -519,6 +519,41 @@ def test_skip_title_heading_abstract(self) -> None:
 
         self.assertEqual(actual, expected)
 
+    def test_unknown_code_language_default(self) -> None:
```
nikaltipar: pytest might be a good replacement for this and similar test suites as it better allows for parameterized tests. (note to self also to check whether it would be a high ROI if we used pytest).

tests/source/unknown_code_language.md
```diff
@@ -0,0 +1,9 @@
+<!-- confluence-page-id: 00000000000 -->
+
+## Unknown language
```
nikaltipar, line 3: I was doing some testing with H2 heading and I saw this getting propagated as document title. I don't remember what the intended behaviour was when front-matter and H1 heading is missing. But I want to highlight this in case we need a bug for this (note to self also to confirm).

md2conf/converter.py
```diff
@@ -806,7 +806,12 @@ def _transform_code_block(self, code: ElementType) -> ElementType:
                 if extension.matches_fenced(language_name, content):
                     return extension.transform_fenced(content)
 
-            language_id = _LANGUAGES.get(language_name)
+            language_id = _LANGUAGES.get(
```
nikaltipar, line 809: If the dict starts getting re-used elsewhere, some light refactoring could be done here, but for now I think this is a good compromise for simplicity.

tests/test_conversion.py
```diff
@@ -519,6 +519,41 @@ def test_skip_title_heading_abstract(self) -> None:
...
+            self.source_dir,
+            self.site_metadata,
+            self.page_metadata,
+        )
+        actual = standardize(doc.xhtml())
+
+        with open(self.target_dir / "unknown_code_language.xml", "r", encoding="utf-8") as f:
+            expected = substitute(self.target_dir, f.read())
+
+        self.assertEqual(actual, expected)
```
hunyadi: This would already be covered by one of the sub-tests in `test_markdown`, wouldn't it? (However, you still need the pass-through test.) This is similar to how `test_admonitions` or `test_broken_links` works: the default case is covered by `test_markdown`, and we have a specific test for the exceptional case.

tests/source/unknown_code_language.md
hunyadi: A unique top-level heading (be it `h1`, `h2` or `h3`) is [promoted to document title](https://github.com/hunyadi/md2conf?tab=readme-ov-file#avoiding-duplicate-titles), and is also removed from the document flow if `skip-title-heading` is enabled. (This is by design.)

tests/test_conversion.py
hunyadi: I agree. However, we have so far had very few parameterized tests. (`test_markdown` has been the notable exception.) `pytest` is much more powerful but the rest of my open-source projects depend on `unittest` (no complex test cases, no extra dependency), and there is undeniably benefit in scale (a.k.a. sharing the same approach across projects). If *md2conf* were the only project I have, I would choose `pytest` over `unittest`.

tests/source/unknown_code_language.md
nikaltipar: ACK, thought it was only H1. My bad. Resolving

tests/test_conversion.py
nikaltipar: Resolving.

tests/test_conversion.py
nikaltipar: Somehow I misread the code, I think this should be OK now

PR #238, opened 2026-03-02
Add recommended minimum version for PlantUML jar

nikaltipar:


README.md
```diff
@@ -62,6 +62,8 @@ npm install -g @mermaid-js/mermaid-cli
    * **Windows**: Download from [graphviz.org](https://graphviz.org/download/)
 3. **Download PlantUML JAR**: Download [plantuml.jar](https://github.com/plantuml/plantuml/releases) and set `PLANTUML_JAR` environment variable to point to it
 
+> The recommended minimum version is **v1.2026.2**; older versions may work but feature support might be limited.
```
hunyadi, line 65: We might also want to update `Dockerfile`, which defines the following:

```dockerfile
ARG PLANTUML_VERSION=1.2025.10
```

`checks.yml` and `integration-tests.yml` run the following command to download the latest version of PlantUML:

```sh
curl -L -o plantuml.jar https://github.com/plantuml/plantuml/releases/latest/download/plantuml.jar
```

We might want to pin this to the same version we are recommending in `README.md`.

nikaltipar: Sure I will do that, thanks for the comment.

PR #266, opened 2026-04-30
Add an extra option to allow user supplied puppeteer configutaion for mmdc command

nenioscio:
Dear md2conf developers,

I open this pull request because the embedded mmdc run failed in our containerized jenkins run. There is alraedy a docker detection in place but it fails in our use-case as the browser install path does not align with our setup.

As a workaround I added an extra option to be passed on the cli so we can add our own puppeteer config file that essentially sets the same args as you but a different browser path.

It would be great to have this merge request accepted so we don't have to maintain the fork and maybe some other users could run with a similar situation.

Alternatively some override parameter could be injected to the mmdc call directly if that makes more sense to you.

Thanks for providing this great tool,

best regards Klaus

### DIFF
--- a/md2conf/mermaid/render.py
+++ b/md2conf/mermaid/render.py
@@ -9,6 +9,7 @@ Copyright 2022-2026, Levente Hunyadi
 import logging
 import os
 import os.path
+import shlex
 from typing import Literal
 
 from md2conf.external import cached_which, execute_subprocess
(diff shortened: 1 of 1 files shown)


md2conf/mermaid/render.py
```diff
@@ -46,7 +46,12 @@ def has_mmdc() -> bool:
     return cached_which(executable) is not None
 
 
-def render_diagram(source: str, output_format: Literal["png", "svg"] = "png", config: MermaidConfigProperties | None = None) -> bytes:
+def render_diagram(
+    source: str,
+    output_format: Literal["png", "svg"] = "png",
+    config: MermaidConfigProperties | None = None,
+    puppeteer_config: str | None = None,
+) -> bytes:
```
hunyadi: Thanks for contributing to *md2conf*! Would you consider something similar to how [PlantUML integration](https://github.com/hunyadi/md2conf/blob/master/md2conf/plantuml/render.py) relies on the environment variable `PLANTUML_CMD`? We had a [similar challenge](https://github.com/hunyadi/md2conf/issues/189) there that we had to configure the theme and include paths but didn't want to add extra parameters to the *md2conf* command line. I believe we could introduce an analogous envrionment variable `MERMAID_CMD`.

nenioscio:
I see you already implemented the change that allows passing the command by environment variable - thanks a lot, I will close the pullrequest

PR #269, opened 2026-05-18
Include generated_by in source digest to detect option changes

karlmolina:
## Problem

When `--generated-by` is changed, pages that were already synced are skipped because the source digest matches — the digest only hashes `converter` options and the raw file bytes, but not `generated_by`.

This means updating the footer message has no effect on existing pages; they are all reported as "Up-to-date page (matching checksum)" and never re-rendered.

## Fix

Include `generated_by` in the MD5 hash so that changing the option invalidates the cached digest and triggers a re-sync of all affected pages.

```python
if self.options.generated_by is not None:
    m.update(self.options.generated_by.encode())
    m.update(b"\n")
```

### DIFF
--- a/md2conf/publisher.py
+++ b/md2conf/publisher.py
@@ -305,6 +305,9 @@ class SynchronizingProcessor(Processor):
         m = hashlib.md5()
         m.update(object_to_json_payload(self.options.converter))
         m.update(b"\n")
+        if self.options.generated_by is not None:
+            m.update(self.options.generated_by.encode())
+            m.update(b"\n")
         m.update(path.read_bytes())
         source_digest = m.hexdigest()
 
--- a/md2conf/publisher.py
+++ b/md2conf/publisher.py
@@ -8,7 +8,7 @@ Copyright 2022-2026, Levente Hunyadi
 
 import hashlib
 import logging
-from dataclasses import dataclass
+from dataclasses import dataclass, fields
 from pathlib import Path
 
 from .api_base import ConfluenceSession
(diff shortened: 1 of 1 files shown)


md2conf/publisher.py
```diff
@@ -305,6 +305,9 @@ def _update_page(self, page_id: ConfluencePageID, document: ConfluenceDocument,
         m = hashlib.md5()
         m.update(object_to_json_payload(self.options.converter))
         m.update(b"\n")
+        if self.options.generated_by is not None:
+            m.update(self.options.generated_by.encode())
+            m.update(b"\n")
```
hunyadi: A minor caveat is that the option `generated_by` can have [templates](https://github.com/hunyadi/md2conf/blob/master/README.md#setting-generated-by-prompt-text-for-wiki-pages), which are substituted at run-time, refer to the function `apply_generated_by_template`. We might want to consider if we want to make a substitution here as well such that different evaluations of the template substitution force a page update.

Another, probably even less critical observation is to choose a separating character that doesn't normally occur in the preceding input. A linefeed (`\n`) never occurs in the output of `object_to_json_payload(self.options.converter)` because any linefeeds are escaped when `orjson` transforms the JSON object. However, `generated_by` is a `str` and can have embedded `\n`. While a minor risk, this allows the "header" (i.e. encoded options & `generated_by`) to "overflow" into the content body. A control character (e.g. NUL byte) might be a safer choice.

Nevertheless, the proposed changes constitute a major leap forward, and cover 95% of the use cases. I am happy to tackle the rest if needed by pushing updates to the pull request.

hunyadi (commented):
That's cool, thanks for spotting and contributing!

md2conf/publisher.py
karlmolina: That would be great. Feel free to update the pull request.

hunyadi (approved):

PR #283, opened 2026-07-16
Detect attachment changes with checksums

### LINKED ISSUE #278: [Bug] Checksums for linked images
### Latest version check

- [x] I confirm I can reproduce the issue with the **latest version** of *md2conf*.

### Describe the issue

I have a GitHub CI/CD action that after release generates pictures for ERDs from mermaid files. These mermaid diagrams are not embedded into the md file because they can be quite heavy and I had to do some custom range-based scaling on file-by-file basis - maybe there is some way how to deal with scaling using md2conf - but not sure if it can be per file. 

For now, I generate PNGs from mermaid files using mermaid CLI and these PNGs are then referenced in md file as image links. What happens now is that md file doesn't technically change because the link didn't change but the picture behind the link changed and page is not updated in Confluence.

Is this an intended behavior or should be locally referenced images also considered when md2conf calculates checksum?

I can work around this issue by adding some logic into GH action if that is intended behavior.

oiahoon:
## Summary

- store a checksum as an attachment content property when using Confluence Cloud REST API v2
- upload same-size attachments when their content changes and skip matching checksums
- retain the existing file-size fallback for Data Center and Server REST API v1

## Testing

- `python -m unittest tests.test_api_v2 tests.test_publisher`
- `python -m unittest discover -s tests`
- `PYTHONPATH=. PYTHON=.venv/bin/python ./check.sh`

Closes #278

### DIFF
--- a/md2conf/api_base.py
+++ b/md2conf/api_base.py
@@ -150,6 +150,36 @@ class ConfluenceSession(ABC):
     @abstractmethod
     def delete_attachment(self, attachment_id: str) -> None: ...
 
+    @property
+    @abstractmethod
+    def supports_attachment_content_properties(self) -> bool:
+        """Whether attachment content properties are available in this Confluence API."""
+
+        ...
+
+    @abstractmethod
+    def get_content_property_for_attachment(self, attachment_id: str, key: str) -> ConfluenceIdentifiedContentProperty | None:
+        """
+        Retrieves a content property for an attachment.
+
+        :param attachment_id: The attachment ID.
+        :param key: The name of the property to fetch (with case-sensitive match).
+        :returns: The content property value, or `None` if not found.
+        """
+
+        ...
+
+    @abstractmethod
+    def update_content_property_for_attachment(self, attachment_id: str, property: ConfluenceContentProperty) -> None:
+        """
+        Creates or updates a content property for an attachment.
+
+        :param attachment_id: The attachment ID.
+        :param property: Content property data to assign.
+        """
+
+        ...
+
     @abstractmethod
     def upload_attachment(
         self,
--- a/md2conf/publisher.py
+++ b/md2conf/publisher.py
@@ -19,7 +19,7 @@ from .collection import ConfluenceUserCollection
 from .compatibility import override, path_relative_to
 from .converter import ConfluenceDocument, apply_generated_by_template, get_orderless_elements, get_volatile_attributes, get_volatile_elements
 from .csf import AC_ATTR, ElementType, elements_from_string
-from .environment import PageError
+from .environment import ArgumentError, PageError
 from .metadata import ConfluencePageMetadata
 from .options import ConfluencePageID, ProcessorOptions
 from .options_converter import ConverterOptions
@@ -404,12 +404,24 @@ class SynchronizingProcessor(Processor):
         base_path = path.parent
         for image_data in document.images:
             name = attachment_name(path_relative_to(image_data.path, base_path))
-            self.api.upload_attachment(page_id, name, attachment_path=image_data.path, comment=image_data.description)
+            self._synchronize_attachment(
+                page_id,
+                attachments.get(name),
+                name,
+                attachment_path=image_data.path,
+                comment=image_data.description,
+            )
             attachments.pop(name, None)
 
         # update attachments with embedded content
         for name, file_data in document.embedded_files.items():
-            self.api.upload_attachment(page_id, name, raw_data=file_data.data, comment=file_data.description)
+            self._synchronize_attachment(
+                page_id,
+                attachments.get(name),
+                name,
+                raw_data=file_data.data,
+                comment=file_data.description,
+            )
             attachments.pop(name, None)
 
         # delete attachments no longer referenced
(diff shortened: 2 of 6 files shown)


tests/test_api_v2.py
hunyadi, line 1: Instead of implementing a brand new mock class, I would extend the existing `MockConfluenceSession` class with the new methods. `MockConfluenceSession` emulates a Confluence instance with a SQLite database.

In my opinion, covering the differences between REST API v1 and REST API v2 is less important. Our main target is REST API v2, and we provide some support for legacy REST API v1. We don't strive for complete feature parity.

tests/test_publisher.py
```diff
@@ -74,6 +75,70 @@ def _get_page_for_document(api: ConfluenceSession, absolute_path: Path) -> Confl
     return api.get_page_properties(props.page_id)
 
 
+class AttachmentCheckingMockConfluenceSession(MockConfluenceSession):
```
hunyadi: Rather than inheriting from `MockConfluenceSession`, I would extend `MockConfluenceSession` with the new methods. `MockConfluenceSession` is expected to implement all the abstract methods declared in `api_base.py`.

tests/test_publisher.py
```diff
@@ -112,6 +177,64 @@ def test_synchronize_directory(self) -> None:
...
+    ) -> tuple[int, int, int]:
+        with _create_temporary_directory() as source_dir:
+            document_path = source_dir / "index.md"
+            attachment_path = source_dir / "diagram.png"
+            document_path.write_text("# Diagram\n\n![diagram](diagram.png)\n", encoding="utf-8")
+            attachment_data = (Path(__file__).parent / "source" / "figure" / "raster.png").read_bytes()
+            attachment_path.write_bytes(attachment_data)
+
+            api = AttachmentCheckingMockConfluenceSession(supports_content_properties=supports_content_properties)
+            try:
```
hunyadi: `MockConfluenceAPI` supports the `with` protocol, and should be preferred over the sequence of `try`, `finally` and `close()`.

md2conf/api_base.py
```diff
@@ -150,6 +150,33 @@ def get_attachment_by_name(self, page_id: str, filename: str) -> ConfluenceAttac
...
+    def get_content_property_for_attachment(self, attachment_id: str, key: str) -> ConfluenceIdentifiedContentProperty | None:
+        """
+        Retrieves a content property for an attachment.
+
+        :param attachment_id: The attachment ID.
+        :param key: The name of the property to fetch (with case-sensitive match).
+        :returns: The content property value, or `None` if not found.
+        """
+
+        raise NotImplementedError("attachment content properties are not supported")
```
hunyadi: Methods in `ConfluenceSession` should all be marked as `@abstractmethod`.

tests/test_publisher.py
oiahoon: Addressed in 1dbe46e. The publisher helper now uses MockConfluenceAPI as a context manager.

tests/test_publisher.py
oiahoon: Addressed in 1dbe46e. MockConfluenceSession now implements the attachment-property methods directly, and the publisher regression uses MockConfluenceAPI.

tests/test_api_v2.py
oiahoon: Addressed in 1dbe46e. Attachment-property coverage now uses the existing SQLite-backed MockConfluenceSession, and the standalone API test file was removed.

md2conf/api_base.py
oiahoon: Addressed in 1dbe46e. The new ConfluenceSession methods are abstract; ConfluenceSessionShared supplies the unsupported fallback, while V2 and the mock override it.

tests/test_publisher.py
```diff
@@ -112,6 +112,57 @@ def test_synchronize_directory(self) -> None:
...
+                    api.remove_content_property_from_attachment(attachment.id, property.id)
+            if change_content:
+                attachment_path.write_bytes(attachment_data[:-1] + bytes([attachment_data[-1] ^ 1]))
+            publisher.process_directory(source_dir)
+            changed_version = api.get_attachment_by_name(page.id, "diagram.png").version.number
+
+            publisher.process_directory(source_dir)
+            unchanged_version = api.get_attachment_by_name(page.id, "diagram.png").version.number
+
+            return first_version, changed_version, unchanged_version
```
hunyadi: I am not sure I understand the return values. `changed_version` and `unchanged_version` should always be equal, shouldn't they? (Reprocessing the folder with no Markdown or attachment changes should not alter the page or attachment version.) We should be able to have the assertion right in the test helper `synchronize_attachment`. If the test helper is not used outside the class `TestPublisher`, we should make it a private function.

Likewise, `changed_version` should always be `first_version + 1`, which we can assert right in this function, we don't need to propagate these to the caller. What do you think?

tests/test_publisher.py
```diff
@@ -112,6 +112,57 @@ def test_synchronize_directory(self) -> None:
 
             publisher.process(sample_dir)
 
+    def synchronize_attachment(
+        self,
+        *,
+        remove_checksum_after_first_upload: bool = False,
+        change_content: bool = True,
```
hunyadi: While I can live with the default values, I would assume the caller always has to pass these parameters to make their intent explicit.

md2conf/publisher.py
```diff
@@ -442,6 +454,55 @@ def _update_page(self, page_id: ConfluencePageID, document: ConfluenceDocument,
...
+        if attachment_path is not None:
+            if not attachment_path.is_file():
+                raise PageError(f"file not found: {attachment_path}")
+            with attachment_path.open("rb") as attachment_file:
+                while chunk := attachment_file.read(1024 * 1024):
+                    digest.update(chunk)
+        elif raw_data is not None:
+            digest.update(raw_data)
+        else:
+            raise NotImplementedError("parameter match not exhaustive")
```
hunyadi: Like with `upload_attachment`, we might want to have a parameter guard that checks if `attachment_path` or `raw_data` is provided but not both:

```python
if attachment_path is None and raw_data is None:
    raise ArgumentError("required: `attachment_path` or `raw_data`")

if attachment_path is not None and raw_data is not None:
    raise ArgumentError("expected: either `attachment_path` or `raw_data`")
```

This prevents callers from getting a `NotImplementedError` when in fact they don't provide the right combination of parameters.

md2conf/api_base.py
```diff
@@ -440,6 +470,19 @@ def _init_site(self, *, domain: str | None, base_path: str | None, space_key: st
...
+    def supports_attachment_content_properties(self) -> bool:
+        return False
+
+    @override
+    def get_content_property_for_attachment(self, attachment_id: str, key: str) -> ConfluenceIdentifiedContentProperty | None:
+        raise NotImplementedError("attachment content properties are not supported")
+
+    @override
+    def update_content_property_for_attachment(self, attachment_id: str, property: ConfluenceContentProperty) -> None:
+        raise NotImplementedError("attachment content properties are not supported")
```
hunyadi: Can you move these to `api_v1.py`? This ensures that implementing these functions is never forgotten, and REST API v2 doesn't accidentally inherit a function that does not actually work. That way, both REST API v1 and REST API v2 implementations must be explicit about what they implement and how, or we get a "compile-time" error with a type checker.

tests/api.py
```diff
@@ -310,6 +310,35 @@ def delete_attachment(self, attachment_id: str) -> None:
...
         self._db.commit()
 
+    @property
+    @override
+    def supports_attachment_content_properties(self) -> bool:
+        return True
+
+    @override
+    def get_content_property_for_attachment(self, attachment_id: str, key: str) -> ConfluenceIdentifiedContentProperty | None:
+        return self.get_content_property_for_page(attachment_id, key)
```
hunyadi: So long as sets of `page_id` and `attachment_id` are disjoint, this is OK. However, I like to be defensive against cryptic errors. If we have separate tables for page content properties and attachment content properties, or a discriminator column the tells which type of object the row applies to, we eliminate the risk of reading a page content property as an attachment content property if `page_id` and `attachment_id` happen to have the same value.

hunyadi (commented):
Thanks for the changes! I have made some additional comments to help make the implementation more robust and easier to maintain. Otherwise, I am happy to merge.

md2conf/publisher.py
oiahoon: Addressed in 9e7dbce. The attachment synchronizer now rejects missing or mutually supplied path and raw-data arguments with ArgumentError.

md2conf/api_base.py
oiahoon: Addressed in 9e7dbce. The unsupported attachment-property implementations now live explicitly on the REST API v1 session.

tests/api.py
oiahoon: Addressed in 9e7dbce. The mock now stores page and attachment content properties in separate tables.

tests/test_publisher.py
oiahoon: Addressed in 9e7dbce. The helper is now private and asserts the expected attachment version transitions internally.

hunyadi (approved):

PR #287, opened 2026-09-01
Include converter version in page digest

### LINKED ISSUE #285: Include converter identity in the page source digest
## Problem

md2conf uses a digest of the Markdown source and conversion options to decide
whether a Confluence page is up to date.

After upgrading md2conf, conversion behavior may change while the Markdown and
options remain unchanged. The existing digest still matches, so the page is
skipped even though the new converter could produce different Confluence Storage
Format output.

This is similar to the option-change invalidation addressed in #269.

## Proposed behavior

Include a converter identity in the page digest so an upgrade can invalidate
previously generated content and trigger synchronization.

Two possible approaches are:

1. Include the package version, causing pages to be synchronized after every
   md2conf release.
2. Introduce an output-format version that is incremented only when conversion
   behavior changes.

Would either approach fit the project's synchronization model? I have a small
implementation and unit test using the package version, but would be happy to
adapt it to the preferred design.

MichelRG74:
The digest that decides whether a page needs to be resynchronized was based only on the Markdown source and conversion options, so upgrading md2conf could leave a previously published page stale until its source changed again. Folding the converter version into the digest makes a release that changes how content is rendered force affected pages to resynchronize on the next run.

Fixes  #285

### DIFF
--- a/md2conf/publisher.py
+++ b/md2conf/publisher.py
@@ -11,6 +11,7 @@ import logging
 from dataclasses import dataclass, fields
 from pathlib import Path
 
+from . import __version__
 from .api_base import ConfluenceSession
 from .api_types import ConfluenceCommentStatus, ConfluenceContentProperty, ConfluenceLabel, ConfluencePage, ConfluenceStatus
 from .attachment import attachment_name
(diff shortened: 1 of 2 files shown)


md2conf/publisher.py
```diff
@@ -139,9 +140,11 @@ class AggregateOptions(ConverterOptions):
     Options that impact the Confluence Storage Format output that is synchronized.
 
     :param generated_by: Text to use as the generated-by prompt (or `None` to omit a prompt).
+    :param converter_version: md2conf version that generated the output.
     """
 
     generated_by: str | None = None
+    converter_version: str = __version__
```
hunyadi: `AggregateOptions` encapsulates options that user controls via command line parameters or a configuration object in Python. However, the user doesn't control the *md2conf* version, it is a constant associated with the package. I understand tests need a configurable `DocumentHasher` such that we can pass arbitrary *md2conf* version numbers. I would suggest we pass the version directly to the initializer of `DocumentHasher` on use rather than indirectly wrapped in `AggregateOptions`.

hunyadi (commented):
Thanks for submitting these changes!

hunyadi (approved):

PR #288, opened 2026-09-01
Handle consecutive GitHub alerts

### LINKED ISSUE #160: Bug in rendering two alert panels one after the other
## Description

When trying to render two alerts block next to each other, the rendering if wrong and only takes into account the first one:

```md
> [!TIP]
> my tip


> [!WARNING]
> My warning whish isn't rendered
```

Image

MichelRG74:
Two `[!TYPE]` alerts written back-to-back with no blank line between them parse as a single Markdown blockquote, so only the first alert was recognized -- the second one's `[!TYPE]` marker leaked into the page as literal text instead of rendering as its own panel. Split such a blockquote into one alert group per `[!TYPE]`-prefixed paragraph, and render each as its own panel or macro.

I would be another way, I guess, to fix #160

Fixes #160

### DIFF
--- a/md2conf/converter.py
+++ b/md2conf/converter.py
@@ -462,10 +462,15 @@ def transform_skip_comments_in_html(html: str) -> str:
 _FOOTNOTE_REF_REGEXP = re.compile(r"^fnref(\d*):(.+)$")
 _TASKLIST_REGEXP = re.compile(r"^\[([x X])\]")
 _GITHUB_ALERT_REGEXP = re.compile(r"^\[!([A-Z]+)\]\s*")
-_GITHUB_ALERT_TO_CSF = {"NOTE": "info", "TIP": "tip", "IMPORTANT": "note", "WARNING": "note", "CAUTION": "warning"}
 
 
-def split_github_alert_groups(root: ElementType) -> None:
+def _is_alert_paragraph(element: ElementType) -> bool:
+    "True if the element is a paragraph that starts with a GitHub alert marker `[!TYPE]`."
+
+    return element.tag == "p" and element.text is not None and _GITHUB_ALERT_REGEXP.match(element.text) is not None
+
+
+def _split_github_alert_groups(root: ElementType) -> None:
     """
     Splits a block-quote with several back-to-back `[!TYPE]` alerts into one block-quote per alert.
 
@@ -481,29 +486,25 @@ def split_github_alert_groups(root: ElementType) -> None:
     :param root: Root of the parsed HTML element tree, mutated in place.
     """
 
-    for blockquote in list(root.iter("blockquote")):
+    for blockquote in list(root.iterchildren("blockquote")):
         children = list(blockquote)
-        if not children or children[0].tag != "p" or not _GITHUB_ALERT_REGEXP.match(children[0].text or ""):
+        if not children or not _is_alert_paragraph(children[0]):
             continue
 
-        group_count = sum(1 for e in children if e.tag == "p" and _GITHUB_ALERT_REGEXP.match(e.text or ""))
+        group_count = sum(1 for e in children if _is_alert_paragraph(e))
         if group_count < 2:
-            continue
-
-        parent = blockquote.getparent()
-        if parent is None:
             continue
 
         groups: list[ElementType] = []
         for element in children:
-            if element.tag == "p" and _GITHUB_ALERT_REGEXP.match(element.text or ""):
+            if _is_alert_paragraph(element):
                 groups.append(blockquote.makeelement(blockquote.tag, blockquote.attrib))
             blockquote.remove(element)
             groups[-1].append(element)
 
-        index = parent.index(blockquote)
+        index = root.index(blockquote)
         groups[-1].tail = blockquote.tail
-        parent[index : index + 1] = groups
+        root[index : index + 1] = groups
 
 
 @dataclass(frozen=True)
@@ -1147,7 +1148,8 @@ class ConfluenceStorageFormatConverter(NodeVisitor):
         if self.options.use_panel:
             return self._transform_panel(blockquote, alert.lower())
         else:
-            class_name = _GITHUB_ALERT_TO_CSF.get(alert)
+            alert_to_csf = {"NOTE": "info", "TIP": "tip", "IMPORTANT": "note", "WARNING": "note", "CAUTION": "warning"}
+            class_name = alert_to_csf.get(alert)
             if class_name is None:
                 raise DocumentError(blockquote, f"unsupported GitHub alert: {alert}")
 
@@ -2009,7 +2011,7 @@ class ConfluenceDocument:
             raise ConversionError(f"failed to convert Markdown file: {path}") from ex
 
         # normalize back-to-back GitHub alerts into separate block-quotes before the tree visitor runs
-        split_github_alert_groups(self.root)
+        _split_github_alert_groups(self.root)
 
         # configure HTML-to-Confluence converter
         converter_options = copy.deepcopy(self.options.converter)
--- a/md2conf/converter.py
+++ b/md2conf/converter.py
@@ -1086,11 +1086,37 @@ class ConfluenceStorageFormatConverter(NodeVisitor):
         if len(blockquote) < 1:
             raise DocumentError(blockquote, "empty GitHub alert")
 
+        pattern = re.compile(r"^\[!([A-Z]+)\]\s*")
+        alert_groups: list[tuple[str, list[ElementType]]] = []
+        for element in blockquote:
+            match = pattern.match(element.text or "") if element.tag == "p" else None
+            if match is not None:
+                alert_element = copy.deepcopy(element)
+                alert_element.text = alert_element.text[len(match.group(0)) :] if alert_element.text is not None else None
+                alert_groups.append((match.group(1), [alert_element]))
+            elif alert_groups:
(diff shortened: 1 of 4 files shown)


md2conf/converter.py
```diff
@@ -1086,11 +1086,37 @@ def _transform_github_alert(self, blockquote: ElementType) -> ElementType:
...
+            if match is not None:
+                alert_element = copy.deepcopy(element)
+                alert_element.text = alert_element.text[len(match.group(0)) :] if alert_element.text is not None else None
+                alert_groups.append((match.group(1), [alert_element]))
+            elif alert_groups:
+                alert_groups[-1][1].append(copy.deepcopy(element))
+
+        if len(alert_groups) > 1:
+            alert_elements: list[ElementType] = []
+            alert_to_csf = {"NOTE": "info", "TIP": "tip", "IMPORTANT": "note", "WARNING": "note", "CAUTION": "warning"}
```
hunyadi: This mapping is already defined below. Let's make sure we promote re-use within a function, and don't introduce duplication.

md2conf/converter.py
```diff
@@ -1086,11 +1086,37 @@ def _transform_github_alert(self, blockquote: ElementType) -> ElementType:
...
+        pattern = re.compile(r"^\[!([A-Z]+)\]\s*")
+        alert_groups: list[tuple[str, list[ElementType]]] = []
+        for element in blockquote:
+            match = pattern.match(element.text or "") if element.tag == "p" else None
+            if match is not None:
+                alert_element = copy.deepcopy(element)
+                alert_element.text = alert_element.text[len(match.group(0)) :] if alert_element.text is not None else None
+                alert_groups.append((match.group(1), [alert_element]))
+            elif alert_groups:
+                alert_groups[-1][1].append(copy.deepcopy(element))
```
hunyadi: I am not sure I understand why we need `deepcopy`. We shouldn't need to copy anything. If we find a `blockquote` with several embedded `p`, each starting with `[!NOTE]` or a similar tag, I believe we should create a new `blockquote` for every group, and *move* existing content over. Then, the existing code can handle the conversion. In other words, we need a preprocessing step.

md2conf/converter.py
```diff
@@ -1086,11 +1086,37 @@ def _transform_github_alert(self, blockquote: ElementType) -> ElementType:
...
+                alert_block[:] = elements
+                if self.options.use_panel:
+                    alert_elements.append(self._transform_panel(alert_block, alert.lower()))
+                else:
+                    class_name = alert_to_csf.get(alert)
+                    if class_name is None:
+                        raise DocumentError(blockquote, f"unsupported GitHub alert: {alert}")
+                    alert_elements.append(self._transform_alert(alert_block, class_name))
+
+            return AC_ELEM("div", {}, *alert_elements)
```
hunyadi: I think there is an opportunity to simplify the code: do a preprocessing step to transform a single `blockquote` with several alert tags into individual `blockquote` elements, and then run the rest of the code on each `blockquote`. This might mean looping over a collection with a single element in most cases.

tests/test_conversion.py
```diff
@@ -201,6 +201,42 @@ def test_admonitions(self) -> None:
...
 
+    def test_github_adjacent_alerts_are_split(self) -> None:
+        content = """<!-- confluence-page-id: 123 -->
+# Test
+
+> [!WARNING]
+> Status: work in progress.
+>
+> This paragraph belongs to the warning.
+>
```
hunyadi: I think it's worth having a test in which this `>` marker is missing, i.e. there are two consecutive alert boxes but they don't have a continuous streak of `>`. Does `blockquote` handling still work in this use case?

tests/test_conversion.py
```diff
@@ -201,6 +201,42 @@ def test_admonitions(self) -> None:
...
+                ProcessorOptions(converter=ConverterOptions(use_panel=True)),
+                self.source_dir,
+                self.site_metadata,
+                self.page_metadata,
+                self.user_metadata,
+            )
+            actual = doc.xhtml()
+            self.assertIn('ac:name="panel"', actual)
+            self.assertIn('panelIconId">26a0-fe0f', actual)
+            self.assertIn('panelIconId">1f4dd', actual)
```
hunyadi: Let's make sure that we don't have magic constants like these GUID fragments, and we tell apart `[!WARNING]` and `[!NOTE]` in some other way.

hunyadi (commented):
Thanks for contributing this feature! I have made a few comments to ensure the implementation remains robust and easy to maintain.

md2conf/converter.py
MichelRG74: Good catch — hoisted `alert_to_csf` into a single module-level `_GITHUB_ALERT_TO_CSF` constant, shared by both code paths. No more duplication. (57f3398)

md2conf/converter.py
MichelRG74: You're right, no copying needed. Added `_split_github_alert_blocks()`: a preprocessing step that walks the original block-quote's children once and *moves* each element into a fresh sibling block-quote (via `blockquote.makeelement(...)`), one per `[!TYPE]` marker found. The existing single-alert logic then runs unchanged on each result. (57f3398)

md2conf/converter.py
MichelRG74: Done — `_transform_github_alert` now just calls `_split_github_alert_blocks()` and loops the existing single-alert conversion (extracted, unchanged, into `_transform_github_alert_block`) over the result, wrapping in `<div>` only when there's more than one. The parallel multi-alert conversion path is gone. (57f3398) Let me know if this matches what you had in mind — happy to adjust further.

tests/test_conversion.py
MichelRG74: Added `test_github_separate_alerts_are_unaffected`, covering two alerts as genuinely separate block-quotes (blank line between them, no shared `>` streak) — confirms that path is untouched by the split logic. (57f3398)

tests/test_conversion.py
MichelRG74: Replaced the hardcoded codepoints with `ConfluencePanel.from_class["warning"].emoji_unicode` / `["note"].emoji_unicode`, so the test derives its expectation from the same mapping the converter uses. (57f3398)

tests/test_conversion.py
```diff
@@ -201,6 +201,83 @@ def test_admonitions(self) -> None:
...
+            )
+            actual = doc.xhtml()
+            self.assertIn('ac:name="panel"', actual)
+            self.assertIn(f'panelIconId">{warning_icon}', actual)
+            self.assertIn(f'panelIconId">{note_icon}', actual)
+            self.assertEqual(actual.count('ac:name="panel"'), 2)
+            self.assertNotIn("[!WARNING]", actual)
+            self.assertNotIn("[!NOTE]", actual)
+        finally:
+            temp_path.unlink(missing_ok=True)
```
hunyadi: Can you fold these tests into `tests/source/alert.md` and `tests/target/alert.xml`? The input/output pairs in `tests/source/*.md` and `tests/target/*.xml` provide stronger guarantees than the string containment tests coded in Python, and are more compact. (Coding agents don't necessarily notice they are there.)

md2conf/converter.py
```diff
@@ -1102,13 +1119,29 @@ def _transform_github_alert(self, blockquote: ElementType) -> ElementType:
...
+            self.visit(e)
+
+        if len(blockquote) < 1:
+            raise DocumentError(blockquote, "empty GitHub alert")
+
+        alert_blocks = self._split_github_alert_blocks(blockquote)
+        if len(alert_blocks) == 1:
+            return self._transform_github_alert_block(alert_blocks[0])
+
+        return AC_ELEM("div", {}, *(self._transform_github_alert_block(alert_block) for alert_block in alert_blocks))
```
hunyadi: I don't think I am comfortable with injecting a `div` that serves no other purpose than to ensure we return a single `ElementType`. This will make the XHTML output for consecutive `[!NOTE]` and `[!TIP]` blocks different based on whether they are separated by nothing (the new algorithm is invoked) or an HTML comment marker (the old algorithm is invoked). In contrast, I believe they should produce identical output.

However, if we move `_split_github_alert_blocks` into `__init__` of `ConfluenceDocument` before we invoke `converter.visit(self.root)`, we can do a pre-processing step on PyMarkdown output before we hand over to the tree visitor. The function `_transform_github_alert` in the tree visitor would then see two separate `blockquote` instances instead of a contiguous block.

tests/test_conversion.py
MichelRG74: Done. Yes very good point, I missed how it was done. So I removed the two Python test methods and folded the scenario into tests/source/alert.md / tests/target/alert.xml as a new "Consecutive alerts" case, alongside the existing single-alert fixtures. Good call, this reads much more compactly. (a9280ec)

md2conf/converter.py
MichelRG74: Agreed, done. I hope I understood your comment correctly. But I moved the split into a `split_github_alert_groups()` preprocessing step called from `ConfluenceDocument.__init__`, before `converter.visit()` runs. `_transform_github_alert` is back to its original single-alert form, no `<div>`. Verified directly: adjacent-alerts-in-one-block-quote and genuinely-separate-block-quotes now produce byte-identical output. (a9280ec)

hunyadi (approved):

PR #290, opened 2026-09-04
Allow a Confluence folder to act as a page parent

alexchernov-web:
### The issue

Confluence Cloud places **folders** in the page hierarchy, and md2conf already models that: `ConfluencePageParentContentType.FOLDER` exists, and a page under a folder deserializes with `parentType: "folder"`.

But it is only ever *parsed*. Nothing can feed a folder in, because every path to a parent goes through `get_page_properties`, and `GET /pages/{id}` returns 404 for a folder ID:

```
404 .../api/v2/pages/1234567890
{"errors":[{"status":404,"code":"NOT_FOUND",
            "title":"Cannot find a page with id [1234567890]"}]}
```

Three call sites reach it, and each is fatal for a tree that contains a folder:

1. `SynchronizingProcessor._synchronize_subtree` — a directory node paired with a folder ID through front matter.
2. `ConfluenceSessionShared.get_or_create_page` — reads the parent only to learn its `spaceId`.
3. `ParentCatalog.is_traceable` — walks a page's ancestors, so it crosses any folder above the page.

The result is that publishing into a space that uses folders dies with an unhandled `HTTPError`. It is not "folders are unsupported" — a folder anywhere above a page makes that page unpublishable.

### The fix

`get_page_properties` falls back to `GET /folders/{id}`.

The fallback goes there rather than at the three call sites because it is the one place all three share, and `is_traceable` in particular walks IDs it never classifies — a narrower fix would have to teach that walk about folders. The cost is that a genuine "no such page" now pays one extra round-trip before it fails.

Two adjustments are needed for the folder payload to fit `ConfluencePageProperties`: `createdAt` is epoch milliseconds where a page uses an ISO-8601 string, and `lastOwnerId` is absent.

### How a folder gets into the tree

Pairing a directory with a folder makes sense together with `synchronized: false`, so md2conf establishes the hierarchy through the directory and never sends it content (a folder has no body):

```markdown
---
page_id: "1234567890"
synchronized: false
---
```

A caller who omits that flag fails loudly on `PUT /pages/{folderId}` rather than corrupting anything.

### Verified against a live instance

Beyond the unit tests, checked against a live Confluence Cloud instance on this branch:

- a new page under a folder-paired directory is created with `parentType: "folder"`;
- an already-bound page under a folder is updated and never re-parented;
- sibling reordering through the v1 move endpoint works with a folder as the parent;
- folders nested inside folders work;
- a `page_id` naming a folder that does not exist fails before anything is written;
- the folder's own version stays at 1 — no body is ever sent to it.

### Test

`tests/test_api_v2.py` covers the fallback without a live instance. `tests/api.py` substitutes the whole session, so `api_v2`'s HTTP layer is not reached by the existing suite — but `_get` is a single seam, and a session double with a canned `_get` is enough to assert that a 404 on `/pages/{id}` routes to `/folders/{id}`, that the folder's epoch-millisecond `createdAt` and absent `lastOwnerId` are normalized, and that any status other than 404 is raised rather than retried against the folder endpoint.

On `master` the first two fail with exactly the error this change is about:

```
requests.exceptions.HTTPError: 404 Client Error
FAILED (errors=2)
```

With the change, the full unit suite passes (135 tests), as do `ruff check`, `ruff format --check` and `mypy md2conf tests`.

Note for a follow-up rather than this PR: `MockConfluenceSession` hardcodes `parentType=ConfluencePageParentContentType.PAGE` and has no notion of folders, so publisher-level behaviour with a folder parent is still only covered by manual testing against a live instance. Glad to extend the mock if you would like that covered.

Reproduced against `master` @ 10a7def; 0.6.2 and 0.6.3 behave the same.

### DIFF
--- a/md2conf/api_v2.py
+++ b/md2conf/api_v2.py
@@ -6,6 +6,7 @@ Copyright 2022-2026, Levente Hunyadi
 :see: https://github.com/hunyadi/md2conf
 """
 
+import datetime
 import logging
 import typing
 from dataclasses import dataclass
@@ -251,8 +252,36 @@ class ConfluenceSessionV2(ConfluenceSessionShared):
 
     @override
     def get_page_properties(self, page_id: str) -> ConfluencePageProperties:
-        path = f"/pages/{page_id}"
-        return self._get(ConfluenceVersion.VERSION_2, path, ConfluencePageProperties)
+        try:
+            path = f"/pages/{page_id}"
+            return self._get(ConfluenceVersion.VERSION_2, path, ConfluencePageProperties)
+        except HTTPError as exc:
+            if exc.response is None or exc.response.status_code != 404:
+                raise
+            return self._get_folder_as_page_properties(page_id)
+
+    def _get_folder_as_page_properties(self, folder_id: str) -> ConfluencePageProperties:
+        """
+        Retrieves a Confluence folder through the page properties type, allowing a folder to act as a page parent.
+
+        Confluence Cloud places folders in the page hierarchy: pages report `parentType: "folder"`, and creating a
+        page accepts a folder ID as `parentId`. However, a folder is not a page, and `GET /pages/{id}` responds with
+        HTTP 404 for a folder ID.
+
+        :param folder_id: The Confluence folder ID.
+        :returns: Folder details in the shape of page properties.
+        """
+
+        path = f"/folders/{folder_id}"
+        data = self._get(ConfluenceVersion.VERSION_2, path, dict[str, JsonType])
+
+        # a folder reports `createdAt` as epoch milliseconds where a page uses an ISO-8601 string, and omits `lastOwnerId`
+        created = data.get("createdAt")
+        if isinstance(created, (int, float)):
+            data["createdAt"] = datetime.datetime.fromtimestamp(created / 1000, datetime.timezone.utc).isoformat()
+        data.setdefault("lastOwnerId", None)
+
+        return json_to_object(ConfluencePageProperties, data)
 
     @override
     def update_page(self, page_id: str, content: str, *, title: str, version: int, message: str) -> None:
--- a/tests/test_api_v2.py
+++ b/tests/test_api_v2.py
@@ -1,0 +1,113 @@
+"""
+Publish Markdown files to Confluence wiki.
+
+Copyright 2022-2026, Levente Hunyadi
+
+:see: https://github.com/hunyadi/md2conf
+"""
+
+import datetime
+import logging
+import unittest
+from typing import Any, TypeVar
+
+from requests import HTTPError, Response
+
+from md2conf.api_types import ConfluencePageProperties, ConfluenceVersion
+from md2conf.api_v2 import ConfluenceSessionV2
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s - %(levelname)s - %(funcName)s [%(lineno)d] - %(message)s",
+)
+
+T = TypeVar("T")
+
+FOLDER_ID = "1234567890"
+
+# as returned by `GET /wiki/api/v2/folders/{id}`: no body, `createdAt` in epoch
+# milliseconds rather than ISO-8601, and no `lastOwnerId`
+FOLDER_PAYLOAD: dict[str, Any] = {
+    "id": FOLDER_ID,
+    "type": "folder",
+    "status": "current",
+    "title": "Planning",
+    "parentId": "1111111111",
+    "parentType": "page",
+    "position": 1781,
+    "authorId": "AUTHOR_ID",
+    "ownerId": "OWNER_ID",
+    "spaceId": "SPACE_ID",
+    "createdAt": 1785247544908,
+    "version": {"number": 1, "minorEdit": False},
+}
+
+
+def _http_error(status_code: int) -> HTTPError:
+    response = Response()
+    response.status_code = status_code
+    return HTTPError(f"{status_code} Client Error", response=response)
+
+
+class StubSession(ConfluenceSessionV2):
+    """
+    A session whose only behavior is a canned `_get`.
+
+    The base class constructor opens a connection to discover the site, which a unit test has no use for; this
+    double supplies the single method `get_page_properties` relies on, and records the paths it was asked for.
+    """
+
+    paths: list[str]
+    _pages_status: int
+
+    def __init__(self, *, pages_status: int) -> None:
+        self.paths = []
+        self._pages_status = pages_status
+
+    def _get(self, version: ConfluenceVersion, path: str, response_type: type[T], *, query: dict[str, str] | None = None) -> T:
+        self.paths.append(path)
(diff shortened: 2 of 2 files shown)


md2conf/api_v2.py
```diff
@@ -251,8 +252,36 @@ def get_page(self, page_id: str) -> ConfluencePage:
...
+        :returns: Folder details in the shape of page properties.
+        """
+
+        path = f"/folders/{folder_id}"
+        data = self._get(ConfluenceVersion.VERSION_2, path, dict[str, JsonType])
+
+        # a folder reports `createdAt` as epoch milliseconds where a page uses an ISO-8601 string, and omits `lastOwnerId`
+        created = data.get("createdAt")
+        if isinstance(created, (int, float)):
+            data["createdAt"] = datetime.datetime.fromtimestamp(created / 1000, datetime.timezone.utc).isoformat()
```
hunyadi, line 281: Strange. [Confluence documentation](https://developer.atlassian.com/cloud/confluence/rest/v2/api-group-folder/#api-folders-id-get) for `createdAt` says:

> Date and time when the folder was created. In format "YYYY-MM-DDTHH:mm:ss.sssZ".

This seems identical to how `createdAt` is returned for regular pages. Is this a Confluence bug?

If `createdAt` can occasionally be an integer, we might want to make the de-serializer more flexible rather than adding behavior specific to folders.

md2conf/api_v2.py
```diff
@@ -251,8 +252,36 @@ def get_page(self, page_id: str) -> ConfluencePage:
...
     def get_page_properties(self, page_id: str) -> ConfluencePageProperties:
-        path = f"/pages/{page_id}"
-        return self._get(ConfluenceVersion.VERSION_2, path, ConfluencePageProperties)
+        try:
+            path = f"/pages/{page_id}"
+            return self._get(ConfluenceVersion.VERSION_2, path, ConfluencePageProperties)
+        except HTTPError as exc:
+            if exc.response is None or exc.response.status_code != 404:
+                raise
+            return self._get_folder_as_page_properties(page_id)
```
hunyadi, line 261: I am afraid I am not fully comfortable with this approach. I agree it solves the immediate problem but it introduces cryptic behavior that may cause a headache to future maintainers of *md2conf*. Once pages and folders are collapsed into the same Python data structure, *md2conf* is unable to tell them apart but pages and folders do behave differently, specifically folders cannot receive content. Furthermore, Confluence REST API has `parentType`, so we can differentiate between pages and folders and create the right object type. *md2conf* never creates folders but maybe it should for directories without `README.md`. (Today, either a `README.md` is always created automatically, or the directory level is skipped.) I would prefer a solution that "teaches" *md2conf* about proper treatment of folders, even if more complex, and doesn't rely on a constellation of front-matter attributes like `synchronized: false` to function properly. Such configuration constellations are difficult to explain to end users in documentation, and the error messages would likely remain unhelpful.

hunyadi (commented):
Thanks for proposing this change! I recognize the problem, and agree that we should update the code such that hierarchies with folders are handled properly. However, I would avoid collapsing different object types in Confluence into the same Python class, which might cause further synchronization problems down the road. Instead, I would add folders as a new Python class, and extend the implementation with the right behavior.

alexchernov-web:
You are right, and I would rather rework this than argue for it.

**The type collapse:** agreed. My own measurements make your case better than my patch did. A folder has no `lastOwnerId` and types `createdAt` differently, so these are two schemas rather than one with a gap. Once they share a class, md2conf can no longer answer "can this receive content?", which is the question that bites later.

**`createdAt`:** this looks like a narrow Confluence bug. Every folder I checked (6, across different spaces, created weeks apart) returns epoch milliseconds, while `version.createdAt` in the same response is correct ISO-8601:

| | `createdAt` | `lastOwnerId` |
|---|---|---|
| `GET /pages/{id}` | `"2026-07-28T14:06:42.471Z"` | present |
| `GET /folders/{id}` | `1785247544908` | absent |

So a more tolerant deserializer is the right answer, as you suggest.

**One thing I should have led with.** The case that motivates this is not the front-matter one. On 0.6.3, `md2conf -r <folder id>` is a plain CLI run with no front matter anywhere, and it dies with a bare `404 .../api/v2/pages/<id>`. With the fix the page is created under the folder. Folders and pages look identical in the Confluence UI that ID gets copied from. The `is_traceable` walk reaches the same 404 for a document whose title happens to match a page living under a folder, again without the user ever thinking about folders. The `page_id` + `synchronized: false` pairing you objected to is the secondary case, and it is the README's own documented pattern. I framed the PR around it, which was a mistake.

One more thing that may shrink the design: the type is almost always known from the API already. `direct-children` returns `type` per child going down, and `parentType` tells you the parent's kind going up. The only ambiguous input is a bare ID from `-r` or front matter, where it has to be probed.

Shall I narrow this PR to the deserializer fix and open a separate issue for the folder design? Glad to implement the folder type if you sketch the shape you want.

hunyadi:
Since this will likely require several back-and-forth trials before we get this right, I have created a new branch `dev`. You might want to target this branch with the pull request.

I have made the following design decisions in the code pushed to the branch `dev`:

* The field `createdAt` has been omitted from folder properties. I could confirm that `createdAt` is indeed a UNIX timestamp of milliseconds since the epoch (transmitted as `int`) but other `createdAt` fields are ISO date-time strings (transmitted as `str`). This makes me suspicious that this is an oversight on behalf of Atlassian, and may be fixed in a future release, which would break compatibility in *md2conf*. Since we don't use `createdAt` for folders, it's best to omit this field.
* `lastOwnerId` has been removed from page properties. We don't currently use this field.
* New API functions like `get_object_space_id` and `get_object_parent_position` have been introduced. This is to ensure compatibility with Confluence REST API v1, which doesn't have a notion of *folder*. They simply route to page properties for API v1.
* `RetryAdapter` has been customized with a new mode to disable retries. New API functions like `get_object_space_id` and `get_object_parent_position` probe various Confluence endpoints, and Confluence returns HTTP 404 both when a page or folder with the specified ID is not found and when the call assumes the wrong type (e.g. calls a page endpoint with a folder ID). We use retries to tackle eventual consistency when a new page is created but keeping retries enabled for probing unnecessarily extends response time.
* Reading folder metadata (e.g. space ID, parent ID or position among siblings) requires specific scopes. If the read scope for folders is missing, *md2conf* still terminates with an error. The extra scopes have been documented in `README.md`.

Given the changes, my tests pass with either the root page set to a folder (instead of a page), or when the content hierarchy contains some folders to group items. Let me know how the new code fares with your setup.

hunyadi:
I am closing this pull request because two main challenges, namely (1) reading the parent only to learn its space ID and (2) tracing the page ancestors, already account for folders, in a way that is compatible with both REST API v1 and v2 (the former of which has no notion of *folder*). Feel free to test the new behavior on the branch *master*, and if you find the implementation inadequate, open an issue or re-open the pull request, rebasing to the latest *master* state. Thank you in advance!

PR #294, opened 2026-09-17
Extend confluence folder support

### LINKED ISSUE #291: [Feature] Allow index.md to represent a Confluence folder
## Motivation

`md2conf` can now use an existing Confluence folder as `--root-page` and can traverse page hierarchies containing existing folders on the current `master` branch. However, it does not allow an `index.md` hierarchy node to represent a Confluence folder.

When `--keep-hierarchy` is enabled and a directory has no `index.md`, `README.md`, or `<directory-name>.md`, md2conf currently creates a synthetic `index.md` containing `[[_LISTING_]]` and publishes it as a Confluence page. A directory with an explicit `index.md` is also always represented as a page.

This means a source hierarchy cannot currently map to a Confluence hierarchy in which intermediate nodes are folders while retaining `index.md` files that provide their proper display titles. Associating an `index.md` with a folder ID does not work because md2conf retrieves and updates it as a page, while a Confluence folder cannot receive a page body.

This has several drawbacks:

1. Structural directories are represented as content pages instead of Confluence folders.
2. Existing Confluence folders cannot be associated with corresponding `index.md` hierarchy nodes in a single synchronization operation.
3. There is no source-controlled way to give a generated folder a friendly display title that differs from the directory name.

The README already lists `write:folder:confluence` and `delete:folder:confluence` as scopes used for directories without an index page, but folder creation does not appear to be implemented.

Related discussion: #290.

## Proposed behavior

Allow an `index.md` to explicitly declare that its directory should be represented as a Confluence folder. For example:

```yaml
---
title: "Product Guides"
content_type: folder
# Optional when binding the descriptor to an existing folder:
page_id: "<folder-id>"
---
```

The exact front-matter property name is open for maintainer feedback.

When directory publishing uses `--keep-hierarchy`:

- An `index.md` explicitly marked as folder metadata causes its directory to be represented by a Confluence folder instead of a page.
- The folder title is read from the existing `title` front-matter property, with the directory name used as a fallback.
- The `index.md` remains in the repository as the source-controlled descriptor for the folder and can store the associated folder ID.
- The `index.md` with `content_type: folder` must be metadata-only because Confluence folders cannot receive page content. That means `[[_LISTING_]]` is unnecessary for folder descriptors because Confluence folders display their children automatically.
- Existing matching folders are reused so repeated synchronization is idempotent.
- Existing page-based behavior remains the default for an `index.md` without the explicit folder declaration.
- `--skip-hierarchy` retains its current flattening behavior.
- An existing page or folder can continue to be supplied as `--root-page`.

For example, this source hierarchy:

```text
docs/
├── index.md                 title: Documentation, content_type: folder
├── guides/
│   ├── index.md             title: Product Guides, content_type: folder
│   ├── getting-started.md
│   └── configuration.md
└── reference/
    ├── index.md             title: Reference, content_type: folder
    └── command-line.md
```

would produce:

```text
Documentation               (folder)
├── Product Guides          (folder)
│   ├── Getting Started     (page)
│   └── Configuration       (page)
└── Reference               (folder)
    └── Command Line        (page)
```

Confluence folder synchronization requires REST API v2. When REST API v1 is selected, an `index.md` explicitly marked as a folder should produce a clear unsupported-feature error rather than silently creating a page. Existing page-based hierarchy behavior remains unchanged for ordinary index documents.

I would be happy to contribute the implementation after confirming the front-matter property name, permitted folder-descriptor body, matching behavior, and REST API v1 behavior.

AndreiShenets:
Closes #291

## Summary

Add support for representing Confluence folders with metadata-only `index.md` descriptors when synchronizing a directory hierarchy with Confluence REST API v2.

A folder descriptor uses `content_type: folder`:

```yaml
---
title: Product Guides
content_type: folder
---
```

On first synchronization, md2conf creates or matches the folder among the intended parent’s direct children and writes the association back to the descriptor:

```markdown
<!-- confluence-folder-id: <folder ID> -->
<!-- confluence-space-key: <space key> -->
```

An explicitly known folder can instead be supplied as front matter:

```yaml
---
content_type: folder
folder_id: "<folder ID>"
---
```

This lets the source-root `index.md` map directly to an existing Confluence folder without requiring `--root-page`.

## Changes

- Add folder support to the synchronization tree and REST API v2 session.
- Add folder creation, direct-child title matching, and explicit folder-ID lookup.
- Support `content_type: folder` descriptors only when `--keep-hierarchy` is enabled.
- Reject folder descriptors with a Markdown body.
- Reject folder synchronization for REST API v1.
- Keep generated folder associations separate from author-managed front matter by writing `confluence-folder-id` HTML comments.
- Preserve explicit `folder_id` front-matter support.
- Prevent a folder ID from being treated as a page ID.
- Make `spaceId` optional when parsing REST v2 direct-child responses, as live responses can omit it even though the documented response model includes it.
- Update README documentation and mock API coverage.

## Validation

- `python -m unittest discover -s tests`
  - 140 passed, 11 skipped
- Ruff lint and format checks
- Mypy for `md2conf`, `tests`, and `integration_tests`
- `git diff --check`

## Manual Confluence Cloud validation

Tested against Confluence REST API v2 using a test folder:

1. Synchronized a documentation test set with a root `index.md` explicitly associated to an existing Confluence folder.
2. Confirmed that the metadata-only descriptors were created as Confluence folders.
3. Confirmed generated folder IDs were written as `confluence-folder-id` comments.
4. Re-ran the identical command without source changes.
5. Confirmed no pages or folders were created, no pages were updated, and attachments were unchanged.


md2conf/api_types.py
```diff
@@ -172,6 +193,38 @@ class ConfluencePageProperties:
     version: ConfluenceContentVersion
 
 
+@dataclass(frozen=True)
+class ConfluenceFolderProperties:
+    """
+    Holds Confluence folder properties used for folder synchronization.
+
+    The `createdAt` field is intentionally omitted because Confluence has returned it in a format inconsistent with
+    other content objects.
```
hunyadi: The original description of this class seems more specific. What is the motivation to update the text?

md2conf/api_v2.py
```diff
@@ -228,6 +205,25 @@ def get_homepage_id(self, space_id: str) -> str:
...
+        return True
+
+    @override
+    def get_object_type(self, object_id: str) -> ConfluenceParentType:
+        data = self._post(
+            ConfluenceVersion.VERSION_2,
+            "/content/convert-ids-to-types",
+            {"contentIds": [object_id]},
+            dict[str, JsonType],
+        )
```
hunyadi: While this is sort of OK, I would prefer we keep track of the content type (e.g. page or folder) of IDs returned by API calls (e.g. in `parentType`) to minimize the total number of API calls we need to make. When synchronizing large trees, as some *md2conf* users do, this can make a difference.

Looking at the API signature, this seems a bulk query endpoint. While it can work, it seems twisted to invoke it for each object ID one by one.

md2conf/api_v2.py
```diff
@@ -292,6 +288,34 @@ def _get_folder_properties(self, folder_id: str, *, retry: bool) -> ConfluenceFo
         path = f"/folders/{folder_id}"
         return self._get(ConfluenceVersion.VERSION_2, path, ConfluenceFolderProperties, retry=retry)
 
+    @override
+    def get_folder_properties(self, folder_id: str) -> ConfluenceFolderProperties:
+        return self._get_folder_properties(folder_id, retry=True)
+
+    @override
+    def get_folder_properties_by_title(self, title: str, *, parent_id: str, parent_type: ConfluenceParentType) -> ConfluenceFolderProperties | None:
```
hunyadi: Since `parent_id` and `parent_type` work in conjunction, we might want to capture then as fields of a data-class, e.g. `ConfluenceTypedID`.

md2conf/api_v2.py
```diff
@@ -292,6 +288,34 @@ def _get_folder_properties(self, folder_id: str, *, retry: bool) -> ConfluenceFo
...
+                path = f"/pages/{parent_id}/direct-children"
+            case ConfluenceParentType.FOLDER:
+                path = f"/folders/{parent_id}/direct-children"
+            case _:
+                raise ConfluenceError(f"unsupported parent type for folder: {parent_type.value}")
+
+        children = json_to_object(list[ConfluenceChildProperties], self._fetch_v2(path))
+        matches = [child for child in children if child.type == ConfluenceParentType.FOLDER and child.title == title]
+        if len(matches) > 1:
+            raise ConfluenceError(f"multiple folders named {title!r} found under parent with ID: {parent_id}")
```
hunyadi: Can this occur in practice? From what I recall, Confluence expects `page` objects to have a unique title, and duplicate titles in the same space are not permitted. Is the `folder` object an exception?

md2conf/local.py
```diff
@@ -55,13 +55,19 @@ def _synchronize_structure(self, tree: DocumentNode) -> dict[str, list[str]]:
         """
 
         for node in tree.all():
-            if node.page_id is not None:
+            if node.folder_id is not None:
+                page_id = node.folder_id
```
hunyadi: Let's make sure we don't call folder IDs by the name `page_id`, even if the original code isn't clear here. If the type of the entity is unknown, let's use a generic name like `object_id`.

md2conf/processor.py
```diff
@@ -35,6 +35,9 @@ class DocumentNode:
 
     absolute_path: Path
     page_id: str | None
+    folder_id: str | None
+    is_folder: bool
+    object_id: str | None
```
hunyadi: This seems confusing to me. Why do we need all three object identifiers (page, folder, neutral) as separate fields? From what I understand, this is a discriminated union: it's either a page ID or a folder ID but never both.

Let's make sure the identifier really works as a discriminated union: if you try to extract a page ID as a folder ID or vice versa, you should get a "loud" exception.

md2conf/processor.py
```diff
@@ -197,6 +207,8 @@ def _check_documents(self, root: DocumentNode) -> None:
         title_to_path: dict[str, Path] = {}
         duplicates: set[Path] = set()
         for node in root.all():
+            if node.is_folder:
+                continue  # unlike page titles, folder titles need not be unique across a space
```
hunyadi: Even if so, I think we might want to stay strict. What happens when a user adds some body to an `index.md`, which needs to be promoted to a page? Suddenly, we need a unique title, and if we don't, synchronization can go amiss in subtle ways.

md2conf/processor.py
```diff
@@ -343,11 +355,29 @@ def _index_file(self, path: Path) -> DocumentNode:
         text = path.read_text(encoding="utf-8")
         document = Scanner().parse(text)
         props = document.properties
-        title = props.title or unique_title(document.text)
+        if props.page_id is not None and props.folder_id is not None:
+            raise PageError(f"expected: only one of `page_id` and `folder_id` in {path}")
+        if props.page_id is not None and props.content_type == "folder":
+            raise PageError(f"expected: `folder_id` instead of `page_id` for folder descriptor: {path}")
+        if props.folder_id is not None and props.content_type == "page":
+            raise PageError(f"expected: `content_type: folder` or no `content_type` when using `folder_id` in {path}")
```
hunyadi, line 360: I believe we should have a discriminated union here to accommodate this logic, possibly captured in a dedicated data-class like `ConfluenceTypedID`.

md2conf/publisher.py
```diff
@@ -218,22 +226,35 @@ def _synchronize_structure(self, tree: DocumentNode) -> dict[str, list[str]]:
...
         topmost_id: str | None = None
-        if tree.page_id is not None:
+        topmost_type: ConfluenceParentType | None = None
+        if tree.is_folder and not self.api.supports_folders:
+            raise PageError(f"Confluence folders require REST API v2 when synchronizing {tree.absolute_path}")
+
+        if tree.folder_id is not None:
+            folder = self.api.get_folder_properties(tree.folder_id)
+            topmost_id = folder.parentId
+            topmost_type = folder.parentType
```
hunyadi: These two work in tandem and should be captured by a separate class.

md2conf/publisher.py
```diff
@@ -347,9 +371,47 @@ def _synchronize_subtree(self, node: DocumentNode, parent_id: ConfluencePageID,
...
+    def _synchronize_folder_subtree(self, node: DocumentNode, parent_id: ConfluencePageID, parent_type: ConfluenceParentType, catalog: ParentCatalog) -> None:
+        """Associates a metadata-only index document with a Confluence folder."""
+
+        if not self.api.supports_folders:
+            raise PageError(f"Confluence folders require REST API v2 when synchronizing {node.absolute_path}")
+
+        if node.folder_id is not None:
+            folder = self.api.get_folder_properties(node.folder_id)
+            catalog.add_known(folder.id)
+            catalog.add_parent(page_id=folder.id, parent_id=folder.parentId, position=folder.position)
```
hunyadi: The catalog might be a good place to keep track of the ID type (page or folder) if we already have this information.

tests/api.py
```diff
@@ -173,20 +195,37 @@ def _row_to_attachment(self, row: sqlite3.Row) -> ConfluenceAttachment:
...
+        parent_id = row["parentId"]
         return ConfluencePageProperties(
             id=row["id"],
             status=ConfluenceStatus.CURRENT,
             title=row["title"],
             spaceId="SPACE_ID",
-            parentId=row["parentId"],
-            parentType=ConfluenceParentType.PAGE,
+            parentId=parent_id,
+            parentType=None if parent_id is None else self.get_object_type(parent_id),
```
hunyadi: This property should come directly from the result-set as this method is intended to be a translator between low-level SQL representation and high-level Confluence model. This function shouldn't call other functions on `self` that trigger other queries. (Data transformers are OK.)

tests/api.py
```diff
@@ -220,15 +259,61 @@ def _row_to_page(self, row: sqlite3.Row) -> ConfluencePage:
     @override
     def get_object_space_id(self, object_id: str) -> str:
         LOGGER.debug("object_id: %s", object_id)
-        return self.get_page_properties(object_id).spaceId
+        return "SPACE_ID"
```
hunyadi: I know this currently evaluates to a constant because all the pages in the test set happen to be in the same space but this may not be so in the future if we extend the test set. Let's keep the future-proof implementation.

README.md
```diff
@@ -137,18 +137,20 @@ We recommend the following scopes for scoped API tokens:
...
 * `write:attachment:confluence`
 * `write:content:confluence`
-* `write:folder:confluence` (only required for directories with no index page)
+* `write:folder:confluence` (only required when creating folders from index descriptors)
 * `write:label:confluence`
 * `write:page:confluence`
 * `delete:attachment:confluence`
 * `delete:content:confluence`
-* `delete:folder:confluence` (only required for directories with no index page)
+* `delete:folder:confluence`
```
hunyadi: This remains an optional permission that you only need if you have a folder specified in your Markdown directory that no longer matches with the Confluence hierarchy. I agree the original description is vague but we need to convey this is not a critical granular permission like `read:page:confluence`. Even if you don't have `delete:folder:confluence`, *md2conf* works for most users.

README.md
```diff
@@ -137,18 +137,20 @@ We recommend the following scopes for scoped API tokens:
 * `read:comment:confluence` (only required for checking comments)
 * `read:content:confluence`
 * `read:content-details:confluence`
+* `read:content.metadata:confluence` (only required when publishing folder descriptors beneath an explicit root)
 * `read:folder:confluence` (only required if your content hierarchy has folders)
+* `read:hierarchical-content:confluence` (only required if your content hierarchy has folders)
```
hunyadi: *md2conf* as per the latest commit on the branch *master* runs nicely on a test site with several folders even though this permission is not granted to the API token. I think we should be more specific as to when you actually need this permission.

tests/test_publisher.py
```diff
@@ -267,6 +291,80 @@ def test_hierarchy(self) -> None:
...
+    def test_folder_descriptor_rejects_markdown_body(self) -> None:
+        with MockConfluenceAPI() as api, _create_temporary_directory() as source_dir:
+            (source_dir / "index.md").write_text("---\ncontent_type: folder\n---\n\nNot allowed.\n", encoding="utf-8")
+            with self.assertRaisesRegex(PageError, "no Markdown body"):
+                Publisher(api, self.get_processor_options(api, keep_hierarchy=True, skip_update=False)).process_directory(source_dir)
+
+    def test_folder_descriptor_rejects_rest_api_v1(self) -> None:
+        with MockConfluenceAPIV1() as api, _create_temporary_directory() as source_dir:
+            (source_dir / "index.md").write_text("---\ncontent_type: folder\n---\n", encoding="utf-8")
+            with self.assertRaisesRegex(PageError, "require REST API v2"):
```
hunyadi: If we are depending on the exact error, we should use a more specific exception type (e.g. `ConfluenceAPIVersionMismatch` that derives from `PageError`) rather than relying on the error string.

tests/api.py
```diff
@@ -220,15 +259,61 @@ def _row_to_page(self, row: sqlite3.Row) -> ConfluencePage:
...
 
     @override
     def get_object_parent_position(self, object_id: str) -> tuple[str | None, int | None]:
-        page = self.get_page_properties(object_id)
-        parent_id = page.parentId
-        position = page.position
+        obj: ConfluenceFolderProperties | ConfluencePageProperties
```
hunyadi, line 278: While this is technically correct, I would think about how we can have a `get_typed_object_parent_position` that takes (and as the first member of the tuple, returns) a `ConfluenceTypedID`, which captures both the object type and its ID. `get_object_parent_position` has been a compromise to introduce folders in a lightweight, mostly transparent manner; functions that don't need to deal with what the object type is can still retrieve and store parent and/or position information. However, we are shifting towards an implementation where folders are a prime object, and have already made many other endpoints distinguished based on object type.

hunyadi (commented):
Thanks for opening the pull request! I have added a few comments that I believe can improve future maintainability. Since many synchronization functions now need understanding if they are dealing with pages or folders, I would recommend that we keep track of object type alongside ID, and enrich relevant API functions that they take and return these typed IDs. This can also help when a folder needs to be promoted to a page or vice versa in response to changes in the Markdown directory tree or the Confluence hierarchy.

md2conf/api_types.py
AndreiShenets: What do you mean with the original description? It is a new type

md2conf/processor.py
AndreiShenets: Although I've removed the check as you asked, it can potentially lead to a case when you are not able to have a folder with the same name but under different parents. I can easily imagine such case.

md2conf/processor.py
AndreiShenets: Have this comment been properly addressed? You highlight the code that relates to the doc metadata analysis. I've added a creation of the discriminated union after that but I have a feeling that you probably wanted something other, I just don't get the idea how discriminated union can help with this.

md2conf/api_types.py
hunyadi: The module `‎md2conf/api_v2.py` has already had this class `ConfluenceFolderProperties`. Previously, folders were not a first-class citizen, and the class definition was private to the REST API v2 implementation. From what it appears, you are moving it to the shared space `‎md2conf/api_types.py` (which is totally fine) but replaced the description text with a more generic reasoning about `createdAt` than what the original class had. I was curious of why you changed the description and kept everything else as is.

md2conf/processor.py
hunyadi: This is fine. I was concerned about the document node (an implementation class, not a data-class for holding local configuration options) having `page_id`, `folder_id` and `object_id` simultaneously. A document node is either a page or a folder, and a type discriminator should tell which.

md2conf/processor.py
```diff
@@ -343,11 +352,36 @@ def _index_file(self, path: Path) -> DocumentNode:
...
+
+        is_folder = self.options.keep_hierarchy and (props.content_type == "folder" or props.folder_id is not None)
+        if props.folder_id is not None and not self.options.keep_hierarchy:
+            raise PageError(f"expected: `--keep-hierarchy` when using `folder_id` in {path}")
```
hunyadi: Minor suggestion: we might want to have a neutral wording as the utility might have been invoked from Python, not the CLI. Python configuration options look like `keep_hierarchy` whereas CLI options have the dashes.

md2conf/processor.py
hunyadi: I totally agree that folders may have recurring names, especially if authors don't use the `title` property in Markdown front-matter. I was trying to think about what happens when an `index.md` that was previously synced to Confluence as a folder gets a body, and now has to be synced as a page. This would mean that we need to delete the folder (no mechanism for this at the moment) and re-create it as a page. Even though the folder title didn't change, it has to now meet the stricter rules for pages with a unique title in the space.

However, we could side-step the problem by saying that you can't promote folders into pages, and if you try, *md2conf* raises an exception with a descriptive error message (not part of this pull request as I saw but we might be able to add this check relatively easily).

hunyadi (commented):
Thanks for addressing the comments! I am going to merge the branch into `dev` such that we can more easily iterate on it, and de-couple it from existing changes on the branch `master` that can be included as part of the `0.6.x` series. This pull request changes the Confluence session API, which some tools rely on, and we also need feedback from users who run on Confluence REST API v1 that it doesn't break anything on their end. As such, I believe first-class folder support might be a flagship feature in the `0.7.x` series.

md2conf/api_types.py
AndreiShenets: ah, got it. No particular reason for that, I just asked an agent to do it, and it changed the message. I will revert.

md2conf/processor.py
AndreiShenets: Then I mark it as resolved.

md2conf/processor.py
AndreiShenets: I would say that such conversions should be forbidden for now. I think about it as an edge case which requires significant logic complication and shouldn't happen too often. Especially taking in account that automatic deletion can be dangerous, for example if I have manually created documents that are not synced automatically. So as you said I wouldn't spend to it time right now. 

The key question should I just keep the strict logic as you suggested initially or allow folder name duplications. I expect that duplicate folder names will come up quite quickly.

AndreiShenets:
> Thanks for addressing the comments! I am going to merge the branch into `dev` such that we can more easily iterate on it, and de-couple it from existing changes on the branch `master` that can be included as part of the `0.6.x` series. This pull request changes the Confluence session API, which some tools rely on, and we also need feedback from users who run on Confluence REST API v1 that it doesn't break anything on their end. As such, I believe first-class folder support might be a flagship feature in the `0.7.x` series.

Two questions:
1. Should I change the target branch to `dev` or would you like to do it yourself?
2. Any idea when you would like to release 0.7.x? No pushing, I am just trying to understand when I am be able to update the image tag and pull the changes to our system.

hunyadi:
I can take care of everything related to approving and merging the pull request into the right branch. However, could you take a look at the [linter output](https://github.com/hunyadi/md2conf/actions/runs/35496915491/job/106041470123)? Currently, the validator `markdown-doc` reports missing documentation on one of the new classes. Make sure that `check.sh`, which runs prior to any release, completes with success with all the `dev` dependencies in `pyproject.toml` installed. Thank you in advance!

Normally, we have a release about once a month. We had the last release in August, and the next is planned for the upcoming week, since we have already a bunch of fixes and improvements accumulated on the branch `master`. Some users pin to the branch `master`, and can give early feedback before a new *md2conf* version is released, which is why we prefer that changes sit on the branch `dev` or `master` a few weeks before we stamp a new version on them. This is especially critical for changes that might impact Confluence REST API v1 integration, for which we don't have a test environment, and rely entirely on community feedback.

AndreiShenets:
Thanks. So as far as I understand the next release will be under 0.6.x umbrella without the new folder functionality. Next release will be under 0.7.x umbrella and with the new folder functionality. Approximately at the end of October.  

I've fixed the CI output. `check.sh` passed for me.

hunyadi:
Be careful, `check.sh` has some optional steps, e.g. it skips validating class and function doc-strings unless `markdown-doc` is installed. However, CI/CD as configured in GitHub Actions, installs all optional development dependencies (including `markdown-doc`) and then runs `check.sh`. Unfortunately, it still [reports a linter error](https://github.com/hunyadi/md2conf/actions/runs/35515892633/job/106091653122). Can you double-check that all linting steps run on your machine, including optional steps near the end of [documentation.py](https://github.com/hunyadi/md2conf/blob/master/util/documentation.py#L175) that rely on `find_spec("markdown_doc")`? Thank you in advance!

AndreiShenets:
yeah, sorry. I think should be fixed now

md2conf/api_v2.py
```diff
@@ -292,6 +269,32 @@ def _get_folder_properties(self, folder_id: str, *, retry: bool) -> ConfluenceFo
...
+                path = f"/pages/{parent_id.page_id}/direct-children"
+            case ConfluenceContentType.FOLDER:
+                path = f"/folders/{parent_id.folder_id}/direct-children"
+            case _:
+                raise ConfluenceError(f"unsupported parent type for folder: {parent_id.type.value}")
+
+        children = json_to_object(list[ConfluenceChildProperties], self._fetch_v2(path))
+        match = next((child for child in children if child.type == ConfluenceContentType.FOLDER and child.title == title), None)
+        if match is None:
+            return None
```
hunyadi: We use the `match` statement and a variable called `match` in the same function body, which may be confusing. Furthermore, a `for` statement with an inner `if` and `break` when the object of the right type and property value is found and an outer `else` otherwise might be easier to read, and is approximately the same lines of code.

md2conf/processor.py
```diff
@@ -343,11 +352,36 @@ def _index_file(self, path: Path) -> DocumentNode:
...
+        if props.folder_id is not None and props.content_type == "page":
+            raise PageError(f"expected: `content_type: folder` or no `content_type` when using `folder_id` in {path}")
+
+        is_folder = self.options.keep_hierarchy and (props.content_type == "folder" or props.folder_id is not None)
+        if props.folder_id is not None and not self.options.keep_hierarchy:
+            raise PageError(f"expected: `keep_hierarchy` enabled when using `folder_id` in {path}")
+        if is_folder:
+            if path.name != "index.md":
+                raise PageError(f"expected: folder descriptor named `index.md`; got: {path}")
+            if document.text.strip():
```
hunyadi: Since we don't use the stripped string later on, a more efficient approach might be to call [str.isspace](https://docs.python.org/3/builtins/stdtypes.html#str.isspace), which elides the copy.

md2conf/publisher.py
```diff
@@ -217,23 +226,29 @@ def _synchronize_structure(self, tree: DocumentNode) -> dict[str, list[str]]:
...
+        topmost_id: ConfluenceTypedID | None = None
+        if tree.is_folder and not self.api.supports_folders:
+            raise ConfluenceAPIVersionMismatch(f"Confluence folders require REST API v2 when synchronizing {tree.absolute_path}")
+
+        if tree.object_id is not None and tree.object_id.type == ConfluenceContentType.FOLDER:
+            folder = self.api.get_folder_properties(tree.object_id.folder_id)
+            topmost_id = folder.parent
+        elif tree.object_id is not None:
+            page = self.api.get_page_properties(tree.object_id.page_id)
+            topmost_id = page.parent
```
hunyadi: Since both of these conditions involve a check on `tree.object_id is not None`, it might be cleaner to structure this as an outer `if` and an inner (exhaustive) `match`, which better conveys the discrimination on `ConfluenceContentType`.

md2conf/publisher.py
```diff
@@ -242,18 +257,17 @@ def _synchronize_order(self, tree: DocumentNode, parent_to_children: dict[str, l
         Recursively arranges child pages of a parent page in the same order as files in their parent directory.
         """
 
-        metadata = self.page_metadata.get(tree.absolute_path)
```
hunyadi, line 245: This check involves calling Confluence REST API to confirm the page (object) exists before attempting to arrange it in some order. This guards against cases when the user associates a Markdown source file with a Confluence page that has been moved to the trash (since the Markdown source file was last edited). If this guarantee is now obtained by some other means, invoking Confluence REST API might be removed.

md2conf/publisher.py
```diff
@@ -242,18 +257,17 @@ def _synchronize_order(self, tree: DocumentNode, parent_to_children: dict[str, l
...
 
         # get order of child pages
         local_order: list[str] = []
         for child in tree.children():
-            metadata = self.page_metadata.get(child.absolute_path)
-            if metadata is None:
+            if child.object_id is None:
                 continue
-            local_order.append(metadata.page_id)
+            if child.object_id.type == ConfluenceContentType.PAGE:
```
hunyadi: Shouldn't ordering apply to all sorts of objects, including pages and folders?

md2conf/publisher.py
```diff
@@ -295,12 +309,16 @@ def _synchronize_users(self, users: set[tuple[str, str]]) -> ConfluenceUserColle
                     user_metadata.add(email, remote_user.accountId)
         return user_metadata
 
-    def _synchronize_subtree(self, node: DocumentNode, parent_id: ConfluencePageID, catalog: ParentCatalog) -> None:
-        if node.page_id is not None:
+    def _synchronize_subtree(self, node: DocumentNode, parent_id: ConfluenceTypedID, catalog: ParentCatalog) -> None:
+        if node.is_folder:
+            self._synchronize_folder_subtree(node, parent_id, catalog)
+            return
```
hunyadi: Since we do an early `return`, and the rest of the statements are about pages only, it might be more readable to have both a `_synchronize_page_subtree` and a `_synchronize_folder_subtree`, and have `_synchronize_subtree` route to the right implementation with a `match` statement. This might be safer if we decide to add other object types, and we need to update the code to support them.

md2conf/publisher.py
```diff
@@ -347,9 +366,47 @@ def _synchronize_subtree(self, node: DocumentNode, parent_id: ConfluencePageID,
...
 
         for child_node in node.children():
-            self._synchronize_subtree(child_node, ConfluencePageID(page.id), catalog)
+            self._synchronize_subtree(child_node, page.typed_id, catalog)
+
+    def _synchronize_folder_subtree(self, node: DocumentNode, parent_id: ConfluenceTypedID, catalog: ParentCatalog) -> None:
+        """Associates a metadata-only index document with a Confluence folder."""
+
+        if not self.api.supports_folders:
+            raise ConfluenceAPIVersionMismatch(f"Confluence folders require REST API v2 when synchronizing {node.absolute_path}")
```
hunyadi: While this is OK, we might want to flag this as early as possible, e.g. when or immediately after the `DocumentNode` tree has been constructed. This prevents partial synchronization when a Confluence REST API v1 use falsely has a `folder_id` in the front-matter properties.

md2conf/api_types.py
```diff
@@ -42,6 +43,60 @@ class ConfluenceParentType(enum.Enum):
...
+    """
+
+    id: str
+    type: ConfluenceContentType
+
+    @property
+    def page_id(self) -> str:
+        """Returns the ID if it identifies a page, and fails otherwise."""
+
+        if self.type != ConfluenceContentType.PAGE:
```
hunyadi: We have a slight preference for using the operators `is` and `is not` in this context when comparing against enumeration value objects. Even though equality comparison also works (and has the same semantics), identity comparison emphasizes that we are really checking if an object *is of* the expected kind, like with type objects. Unfortunately, *md2conf* hasn't had discriminated unions so far and there is no clear established practice in this very project, other projects of ours have the identity comparison for enumeration values, and may help our maintainers on the long run to use the same syntax.

hunyadi (approved):
Thanks for the changes in the latest iteration! I have done another review pass and I have only found minor issues. If you have the time and willing to act upon the suggestions, I would definitely appreciate, otherwise I am going to take care of the comments when I merge into `dev` or `master`.

md2conf/publisher.py
AndreiShenets: I haven't changed anything here, the agent explanation is:

> I did not make a new code change specifically for that comment, because the existence guarantee is already preserved earlier in the flow:
> 
> 1. Processor.process_directory() completes _synchronize_structure() before _synchronize_content().
> 2. _synchronize_page_subtree() calls get_page_properties() for explicitly associated pages, verifying the remote page exists.
> 3. _synchronize_folder_subtree() similarly calls get_folder_properties().
> 4. Only afterward does _synchronize_content() invoke _synchronize_order().
> 
> Therefore, by the time ordering reads tree.object_id, the corresponding remote object has already been successfully fetched or created. A missing/trashed object will fail during structure synchronization rather than ordering.
> 
> So this review comment is addressed by confirming the replacement guarantee, not by restoring the removed page_metadata lookup. This reasoning should be included in the PR thread response.

md2conf/publisher.py
AndreiShenets: No, Confluence API doesn't support the folder ordering. I put a comment regarding this.

AndreiShenets:
I hope I've addressed all your comments correctly 🙂

